PR #14, opened 2016-04-01
chore(common): Only keep 10 builds

jchauncey:
There is no reason to keep around 42 builds. 10 should be sufficient

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -19,7 +19,7 @@ repos.each { Map repo ->
 }
 
 defaults = [
-  numBuildsToKeep: 42,
+  numBuildsToKeep: 10,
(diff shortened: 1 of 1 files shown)


mboersma:
I have often had need to read old job logs, because IRL sometimes it isn't possible to revisit a fix or proposal for days or even weeks. 42 was quirky, but honestly I'd prefer a larger value, like 100 that the venerable [test-ec2](https://ci.deis.io/job/test-ec2/) job uses.

mboersma:
Does anyone else feel strongly about increasing or decreasing the default # of Jenkins builds to retain? Otherwise we should leave it as-is.

vdice:
I'd strongly opt _not_ to decrease the number while we're still in the early stages of making test runs more deterministic.  Therefore, I'd vote to keep as-is -- or _increase_ if there was a strong opinion for the latter.

PR #27, opened 2016-04-04
fix(workflow_test): remove hard-coded slack integration token

vdice:
With plans to revisit once plugin supports secure credentials handling:
https://github.com/jenkinsci/slack-plugin/pull/208

Background on need for this functionality: https://github.com/jenkinsci/slack-plugin/issues/195

With this change, slack notifications will revert to using the global defaults (meaning will only be able to broadcast to the one default channel at this time.)

cc @Joshua-Anderson


jchauncey:
🚢

PR #34, opened 2016-04-07
feat(jobs): add curlStatus method to report build status to GitHub

vdice:
Also remove blocking dependency between upstream jobs
and downstream test job(s)


vdice:
Example of this DSL tested lives in https://ci.deis.io/job/workflow-e2e-pr-clone-job/36/console (job will be deleted after this PR) and test PR [here](https://github.com/deis/workflow-e2e/pull/154) (there are a handful of 'test' contexts/statuses there; the 'final' version uses context `ci/jenkins`)

PR #38, opened 2016-04-07
feat(jobs): adds ghprb plugin commitStatus defaults...

vdice:
Since it will update commit status regardless. :(

### DIFF
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -1,5 +1,7 @@
 evaluate(new File("${WORKSPACE}/common.groovy"))
 
+import utilities.StatusUpdater
+
 repos.each { Map repo ->
   [
     // for each repo, create a <repo.name>-master and <repo.name>-pr job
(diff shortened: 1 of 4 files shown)


jobs/component_jobs.groovy
```diff
@@ -87,15 +98,22 @@ repos.each { Map repo ->
...
+          postBuildScripts {
+            onlyIfBuildSucceeds(false)
+            steps {
+              statuses.each { buildStatus, commitStatus ->
+                conditionalSteps {
+                  condition {
+                    status(buildStatus, buildStatus)
+                    steps {
+                      shell(
+                        new GroovyShell().parse(new File("${WORKSPACE}/common.groovy")).with {
```
vdice: Was hitting errors attempting to invoke a method defined in a diff groovy script (`common.groovy` in this case).  Turns out that is a groovy thang.  http://stackoverflow.com/questions/9004303/load-script-from-groovy-script

jobs/component_jobs.groovy
```diff
@@ -87,15 +100,19 @@ repos.each { Map repo ->
...
+          def statuses = [['SUCCESS', 'pending'],['FAILURE', 'failure'],['ABORTED', 'error']]
+          postBuildScripts {
+            onlyIfBuildSucceeds(false)
+            steps {
+              statuses.each { buildStatus, commitStatus ->
+                conditionalSteps {
+                  condition {
+                    status(buildStatus, buildStatus)
+                    steps {
+                      shell StatusUpdater.curlStatus(
```
sgoings: I don't think this will do what you expect it to do because everything in the `steps { shell ...` block will be sent to be held within the Jenkins job config and executed when the job is run. In other words, `StatusUpdater.curlStatus(...` will be called in bash on the Jenkins slave when the job executes.

vdice: I will say that I've verified `shell curlStatus(args…)` does work as intended (when `curlStatus` declaration is in same file) b/c that method just returns the string (comprising the shell script to execute) to the `shell` closure.  So, the only addition here is attempting to access `curlStatus` from a different file via the class it is declared in, `StatusUpdater`... but yes, still learning the nuances of Groovy

sgoings: Yeah - I totally spaced that part of the operation. I think you're good with this route.

jchauncey:
🚢

PR #42, opened 2016-04-08
Revert "feat(component_jobs): this removes the use of the ghprb plugin"

vdice:
This reverts commit aa3e891668ad563f1ae9325c7acf06722da93c67.

PR #43, opened 2016-04-10
fix(component_jobs): the commit needed to update GH PR status is ghprbActualCommit, not GIT_COMMIT

vdice:
As the two are different, the previous status updates were going to the
wrong status endpoint (GIT_COMMIT, or local merge commit).  

This is now fixed, using the actual/initial commit SHA offered by the
ghprb plugin in the form of the 'ghprbActualCommit' env var.

### DIFF
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -51,6 +51,31 @@ repos.each { Map repo ->
         authenticationToken('5ISZbTayHuC0nHV')
       }
 
+      if (isPR) { // set up GitHubPullRequest build trigger
+        triggers {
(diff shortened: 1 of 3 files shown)


jobs/component_jobs.groovy
```diff
@@ -51,6 +51,31 @@ repos.each { Map repo ->
         authenticationToken('5ISZbTayHuC0nHV')
       }
 
+      if (isPR) { // set up GitHubPullRequest build trigger
```
vdice, line 54: this identical code change is also located in the revert PR at https://github.com/deis/jenkins-jobs/pull/42

PR #47, opened 2016-04-13
feat(release): add release test job

sgoings:


vdice:
@sgoings we now have a `workflow-test-release` as seen here: https://github.com/deis/jenkins-jobs/blob/master/jobs/workflow_test_release.groovy 

If this suits our needs at this time, I propose we close this ticket and file a new one around any suggested amendments/improvements against existing DSL.

sgoings:
Seems 👍 to me, @vdice!

PR #56, opened 2016-04-19
feat(component_job): build and deploy regardless if master or PR

vdice:


jobs/component_jobs.groovy
```diff
@@ -104,22 +101,20 @@ repos.each { Map repo ->
...
+          set -eo pipefail
 
-            make bootstrap || true
+          make bootstrap || true
+
+          export IMAGE_PREFIX=deisci
+          docker login -e="$DOCKER_EMAIL" -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
+          DEIS_REGISTRY='' make docker-build docker-push
+          docker login -e="$QUAY_EMAIL" -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
+          DEIS_REGISTRY=quay.io/ make docker-build docker-push
```
vdice: note this would push both immutable and mutable tags both on PR job and master job -- do we want logic to only push immutable on PR and then both on master?

krancour: > do we want logic to only push immutable on PR and then both on master?

I think so, yes.

PR #58, opened 2016-04-20
fix(component_jobs): grab sha1 as actual commit if job run manually

vdice:
cc @jchauncey


jchauncey:
we should get this merged asap since using the sha to kick off a pr job is a legitimate workflow

jobs/component_jobs.groovy
```diff
@@ -120,7 +127,7 @@ repos.each { Map repo ->
                   'UPSTREAM_BUILD_URL': '${BUILD_URL}',
                   'UPSTREAM_SLACK_CHANNEL': "${repo.slackChannel}",
                   'COMPONENT_REPO': "${repo.name}",
-                  'ACTUAL_COMMIT': '${ghprbActualCommit}',
+                  'ACTUAL_COMMIT': '${ACTUAL_COMMIT}',
```
vdice: this is not quite right yet... the export in shell step above doesn't seem to persist when we invoke here

vdice:
Latest iteration has been tested w/ the manual trigger case; looking good.

PR #60, opened 2016-04-20
chore(all jobs): Increase kept builds to 14 days (one sprint)

krancour:
All roads lead to this job... it's being run so many times a day that if you click "details" in your PR with the intention of examining or restarting a failed job, chances are quite high that failed execution has already rotated out of the kept builds.

**EDIT:**

~~This PR increases the number of kept builds, specifically for the `workflow-test-pr` job, from the default of 42 up to (an arbitrarily chosen and hopefully high enough) 250.~~

This PR makes jobs keep builds for 14 days-- which is the current length of a single sprint.  Unless we're running late, this should allow us to restart any job from the current spring up to the last day of that sprint.

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -21,7 +21,7 @@ repos.each { Map repo ->
 TEST_JOB_ROOT_NAME = 'workflow-test'
 
 defaults = [
-  numBuildsToKeep: 42,
+  daysToKeep: 14,
   bumpverCommitCmd: 'git commit -a -m "chore(versions): ci bumped versions via ${BUILD_URL}" || true',
   testJob: [master: "${TEST_JOB_ROOT_NAME}", pr: "${TEST_JOB_ROOT_NAME}-pr"],
   maxBuildsPerNode: 1,
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -44,7 +44,7 @@ repos.each { Map repo ->
       }
 
       logRotator {
-        numToKeep defaults.numBuildsToKeep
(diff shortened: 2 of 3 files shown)


jchauncey:
added LGTM but just make every job keep 250 for consistency

krancour:
@jchauncey I'm ok with that, but do we think that's too heavy handed?  I'll go with whatever the consensus is on that.

jchauncey:
instead of just a number we could age the builds off by days (which i find better honestly) and keep like 5 days worth of builds

krancour:
@jchauncey let me see if I can figure out how to do that with the DSL.  I think I'd prefer a 14 day window.  That aligns with our current sprint length.  This way, unless we're running late, we can restart any job from the current sprint up to the day we release.

krancour:
Removing "awaiting review" because I'm going to try and make this time-based instead as @jchauncey has suggested.

vdice:
good call @jchauncey, I like that better as well.  @krancour you're probably here already but: https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.jobs.FreeStyleJob.logRotator

jchauncey:
14 days is good

jchauncey:
can you change the PR message (not the commit message) to reflect the change?

krancour:
@jchauncey I did, actually.

krancour:
Oh... nevermind.  I read you wrong.  Fixing.

jchauncey:
=)

krancour:
All set.

PR #67, opened 2016-04-28
chore: Add stdout-metrics to jobs list

jchauncey:


mboersma:
LGTM (but I think the #metrics slack channel still needs to be born).

PR #68, opened 2016-04-28
feat(workflow_test_release): add workflow_test_release job DSL

vdice:
To be used for testing a workflow chart associated with a given release.

TODO after LGTMs:
- [x] update https://ci.deis.io/job/workflow-test-seed-job/ to build this job as well

### DIFF
--- a/jobs/workflow_test.groovy
+++ b/jobs/workflow_test.groovy
@@ -12,7 +12,7 @@ import utilities.StatusUpdater
   name = defaults.testJob[config.type]
   repoName = 'charts'
 
-  testReportMsg = "Test Report: ${JENKINS_URL}job/${name}/\${BUILD_NUMBER}/testReport"
+  testReportMsg = defaults.testJob["reportMsg"]
   upstreamJobMsg = "Upstream job: \${UPSTREAM_BUILD_URL}"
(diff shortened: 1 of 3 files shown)


vdice:
Live example of this job exists in https://ci.deis.io/job/workflow-test-release/2/ which was built straight from the DSL on this branch...

jobs/workflow_test.groovy
```diff
@@ -131,7 +129,6 @@ import utilities.StatusUpdater
 
     environmentVariables {
       env('COMMIT', isMaster)
-      env('PARALLEL_TESTS', true)
```
bacongobbler, line 134: how come we're removing the parallel tests envvar here?

vdice, line 134: Good question.  This was a temporary env var used when we had two e2e charts, one for running tests in a serialized fashion, the other concurrently.  We now just have one e2e chart, defaulting to running in parallel (using multiple `GINKGO_NODES`).  To run in serial using the same chart, we can set `GINKGO_NODES` to `1`.

bacongobbler, line 134: Right! I forgot about that.

Issue #71, opened 2016-04-29
create seed-repo job


bacongobbler:
This job will apply [seed-repo](/deis/seed-repo) to any component under our control. This will help with applying any changes made to seed-repo immediately without any user intervention.

Usage should be:
- make a PR on seed-repo
- PR gets reviewed and approved
- PR gets merged, kicking off a seed-repo job
- all existing components under the seed-repo umbrella are updated

related to work at #55 for automating release tasks.

bacongobbler:
this is partially complete. We have a seed-repo job, but it isn't running due to failures installing the deps. Following up in https://github.com/deis/seed-repo/issues/14

PR #74, opened 2016-04-29
feat(jobs): add seed-repo job

bacongobbler:
This job will run deis/seed-repo against all of the workflow components on any merge to
deis/seed-repo. This allows all of our components to be kept up-to-date by making changes in one
place, and not having us to manually run this against all of our repositories.

This is my first PR against jenkins-jobs so please let me know if I messed something up. I verified this against the Jenkins Job DSL Playground and it generated the XML I desired (after directly importing common.groovy into the script and removing things like WORKSPACE and JENKINS_URL)

addresses #71

### DIFF
--- a/jobs/seed_repos.groovy
+++ b/jobs/seed_repos.groovy
@@ -1,0 +1,53 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+import utilities.StatusUpdater
+
+name = 'deis-seed-repos'
+repoName = 'seed-repo'
+
+job(name) {
+  description """
+    <p>Runs deis/seed-repo against all of the workflow components.</p>
+  """.stripIndent().trim()
+
+  scm {
+    git {
+      remote {
+        github("deis/${repoName}")
+      }
+      branch('master')
+    }
+  }
+
+  logRotator {
+    daysToKeep defaults.daysToKeep
+  }
+
+  triggers {
+    githubPush()
+  }
+
+  wrappers {
+    timeout {
+      absolute(defaults.testJob["timeoutMins"])
+      failBuild()
+    }
+    timestamps()
+    colorizeOutput 'xterm'
+    credentialsBinding {
+      string("GITHUB_ACCESS_TOKEN", "8e11254f-44f3-4ddd-bf98-2cabcb7434cd")
+    }
+  }
+
+  steps {
+    repos.each { Map repo ->
+      shell """
+      #!/usr/bin/env bash
+
+      set -eo pipefail
+
+      ./seed-repo deis/${repo.name}
+      """.stripIndent().trim()
+    }
+  }
+}
--- a/jobs/seed_repos.groovy
+++ b/jobs/seed_repos.groovy
@@ -36,14 +36,14 @@ job(name) {
   }
 
   steps {
-    repos.each { Map repo ->
-      shell """
-      #!/usr/bin/env bash
+    shell """
+    #!/usr/bin/env bash
 
-      set -eo pipefail
+    set -eo pipefail
 
-      ./seed-repo deis/${repo.name}
-      """.stripIndent().trim()
-    }
+    bundle install
+
+    for i in $(./list-repos); do ./seed-repo $i; done
+    """.stripIndent().trim()
   }
 }
--- a/jobs/seed_repos.groovy
+++ b/jobs/seed_repos.groovy
@@ -1,7 +1,5 @@
 evaluate(new File("${WORKSPACE}/common.groovy"))
 
-import utilities.StatusUpdater
-
 name = 'deis-seed-repos'
 repoName = 'seed-repo'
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Also note that this will require us to install ~~the `octokit` gem as well as~~ ruby and bundler on all of our jenkins worker nodes since seed-repo is a bunch of ruby scripts.

jobs/seed_repos.groovy
```diff
@@ -0,0 +1,53 @@
...
+  logRotator {
+    daysToKeep defaults.daysToKeep
+  }
+
+  triggers {
+    githubPush()
+  }
+
+  wrappers {
+    timeout {
```
vdice: we probably don't need a `timeout` closure for this job as it _should_ be pretty quick...

jobs/seed_repos.groovy
```diff
@@ -0,0 +1,53 @@
...
+      branch('master')
+    }
+  }
+
+  logRotator {
+    daysToKeep defaults.daysToKeep
+  }
+
+  triggers {
+    githubPush()
```
vdice, line 25: This will trigger the job on any commit to the master branch of the `seed-repo` repo -- could be what we had in mind, otherwise, we can remove this if we're thinking this job will be manually kicked off for a given repo.

bacongobbler, line 25: no I wanted this job to be triggered on any commit to seed-repo so the repositories are up to date on any changes made there

bacongobbler, line 25: for example merging https://github.com/deis/seed-repo/pull/11 would now close the beta3 milestone on every repository, helping with release automation.

jobs/seed_repos.groovy
bacongobbler: (famous last words)

I'll remove this. I agree it should be quick :)

jobs/seed_repos.groovy
```diff
@@ -0,0 +1,53 @@
...
+    }
+    timestamps()
+    colorizeOutput 'xterm'
+    credentialsBinding {
+      string("GITHUB_ACCESS_TOKEN", "8e11254f-44f3-4ddd-bf98-2cabcb7434cd")
+    }
+  }
+
+  steps {
+    repos.each { Map repo ->
```
vdice: This looks good -- the only problem I see (not at all part of this code, just sayin') is that we have a list of `repos` defined in `common.groovy` but which _may_ diverge from the repos that `seed-repo` tracks... I'd be inclined to use the list of repos (if exists) in `seed-repo` for iterating/seeding, if possible.

vdice:
@bacongobbler if the jenkins nodes/slaves already have ruby (not sure if they do) then we should be able to just call `bundle install` in the execute shell portion of this job before iterating thru repos and seeding, right?

jobs/seed_repos.groovy
vdice, line 25: Nice; that does make sense!

jobs/seed_repos.groovy
bacongobbler: good thinking. I completely forgot about `list-repos`.

bacongobbler:
Yes, we should be able to. :)

arschles:
Very cool patch. Is this something that we shouldn't be doing on a travis job in seed-repo though? Does it matter where we do it?

bacongobbler:
@arschles it doesn't really matter where we do it. Travis isn't hooked up to seed-repo and I did this more as an experiment to play around with our jenkins DSL jobs, so there's my justification for it being here vs. travis :)

arschles:
@bacongobbler sgtm. good luck!

jobs/seed_repos.groovy
```diff
@@ -0,0 +1,49 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+import utilities.StatusUpdater
```
vdice: We won't need this import (only needed for PR jobs that update commit statuses)

vdice:
Looks good @bacongobbler !  The only manual step left is to either modify one of our existing seed jobs -- probably https://ci.deis.io/job/component-seed-job/ -- to track any changes in this DSL and rebuild the job if necessary (see `Process Job DSLs` section of that job for examples)

bacongobbler:
done! component-seed-job is now tracking seed_repos.groovy. I put it under monitor_jobs.groovy. Is there a reason why the component_jobs, workflow_test and workflow_test_release DSL jobs are missing from that list right now?

vdice:
`component_jobs.groovy` is the first block (or should be) -- it didn't show up immediately for some reason when I last looked.  The `*test` jobs are currently tracked separately by https://ci.deis.io/job/workflow-test-seed-job/ .

jchauncey:
we should have just 1 seed job since we only have 1 repo. Will make things easier eventually

vdice:
@jchauncey I agree (and perhaps @bacongobbler does as well) It is the direction we want to go for sure.

bacongobbler:
it would be nice, but I like to :ship: whatever works, document it and improve/bikeshed later. :bike:

Issue #80, opened 2016-05-04
component-seed-job failing


bacongobbler:
The component-seed-repo job is now failing as of https://github.com/deis/jenkins-jobs/commit/e30a74f7892c5482ca441c24156a86a90a2451d1. See https://ci.deis.io/job/component-seed-job/56/console

```
09:45:17 Processing DSL script monitor_jobs.groovy
09:45:17 ERROR: (monitor_jobs.groovy, line 43) No such property: isPr for class: javaposse.jobdsl.dsl.helpers.scm.GitContext
09:45:17 [BFA] Scanning build for known causes...
09:45:17 [BFA] No failure causes found
09:45:17 [BFA] Done. 0s
09:45:17 Finished: FAILURE
```

bacongobbler:
ping @jchauncey

bacongobbler:
The [job has gone green again](https://ci.deis.io/job/component-seed-job/57/console) after #81.

PR #81, opened 2016-05-04
fix(monitor): Just remove inclusion check altogether

jchauncey:


jchauncey:
fixes #80

bacongobbler:
Is there a reason that this was originally needed? I was trying to hack up a PR to make it work, but if there's no need for this then we can merge, but I'd be happy to take the time and find a fix.

jchauncey:
Yes the monitoring repo contains 3 sub dirs and I would rather do an inclusion check to see which one to build. Right now this doesnt work with the github pr builder but figured the master jobs could continue to use it. Instead, we will just build all 3 on new master run

bacongobbler:
Excellent. [The job's green again](https://ci.deis.io/job/component-seed-job/57/console). Thank you!

PR #85, opened 2016-05-18
feat(storage): add storage support for e2e tests

smothiki:
GCS backend

### DIFF
--- a/jobs/storage_test.groovy
+++ b/jobs/storage_test.groovy
@@ -1,0 +1,67 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+name = "storage_backend_e2e"
+repoName = 'charts'
+
+job(name) {
+  description """
+    <p>Nightly Job runs the <a href="https://github.com/deis/workflow-e2e">e2e tests</a> against a <a href="https://github.com/deis/charts/tree/master/${defaults.workflowChart}">${defaults.workflowChart}</a> chart configured to GCS by default </p>
+  """.stripIndent().trim()
+
+  scm {
+    git {
+      remote {
+        github("deis/${repoName}")
+      }
+      branch('master')
+    }
+  }
+
+  logRotator {
+    daysToKeep defaults.daysToKeep
+  }
+  parameters {
+   stringParam('STORAGE_TYPE', 'gcs', "storage backend for helm chart, default is gcs")
+  }
+
+  triggers {
+    cron('@daily')
+  }
+
+  wrappers {
+   buildName('Storage_backed-${STORAGE_TYPE} #${BUILD_NUMBER}')
+    timeout {
+      absolute(30)
+      failBuild()
+    }
+    timestamps()
+    colorizeOutput 'xterm'
+    credentialsBinding {
+      string("GCLOUD_CREDENTIALS", "246d6550-569b-4925-8cda-e11a4f0d6803")
+      string("GCS_KEY_JSON", "71dd890b-e4ce-4483-97ca-a7c286c2381c")
+      string("AWS_SECRET_KEY","033f76e3-21b8-4bae-a7b3-7f43e07138d3")
+      string("AWS_ACCESS_KEY","02a5c84e-3248-4776-acc5-6b6a1fb24dfc")
+    }
+  }
+
+  steps {
+    shell '''
+      #!/usr/bin/env bash
+
+      set -eo pipefail
+
+      if [ ${STORAGE_TYPE} == "gcs" ]; then
+        TYPE="GCS"
+      else
+        TYPE="AWS"
+        export S3_REGION="us-west-1"
+      fi
+      export ${TYPE}_REGISTRY_BUCKET="deis-registry-${BUILD_NUMBER}"
+      export ${TYPE}_BUILDER_BUCKET="deis-builder-${BUILD_NUMBER}"
+      export ${TYPE}_DATABASE_BUCKET="deis-database-${BUILD_NUMBER}"
+
+      ./ci.sh
+    '''.stripIndent().trim()
+
+  }
+}


smothiki:
e2e running with storage backend configured to GCS
https://ci.deis.io/job/storage_backend/6/console

smothiki:
Once merged requires a change in https://ci.deis.io/job/recoverability_seed_job/ configuration to raplace smothiki with deis

jobs/storage_test.groovy
```diff
@@ -0,0 +1,57 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+name = "storage_backend"
+repoName = 'charts'
+
+job(name) {
+  description """
+    <p>Runs the <a href="https://github.com/deis/workflow-e2e">e2e tests</a> against a <a href="https://github.com/deis/charts/tree/master/${defaults.workflowChart}">${defaults.workflowChart}</a> chart</p>
```
vdice: maybe just add a mention about the particular use of this job, which is to run existing e2e tests but using a non-default storage backend...

jobs/storage_test.groovy
```diff
@@ -0,0 +1,57 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+name = "storage_backend"
```
vdice: this job currently hardcodes `gcs` as the storage backend... maybe we rename this job to `gcs_storage_backend_e2e` -- unless there are plans for this one job to test multiple storage backends...

jobs/storage_test.groovy
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  steps {
+    shell '''
+      #!/usr/bin/env bash
+
+      set -eo pipefail
+
+      export STORAGE_TYPE="gcs"
```
vdice: It could be handy to leave this a generic `storage_backend_e2e` job and expose `STORAGE_TYPE` as a parameter that can be supplied when the job is run (defaulting to `gcs` or some other default if not supplied).  then, this same job dsl/config could be used to run against all needed storage backends...

If this might be a good idea, we can create a ticket to work on at a later date; not needed for this PR, just a thought.

vdice:
@smothiki 

> Once merged requires a change in https://ci.deis.io/job/recoverability_seed_job/ configuration to raplace smothiki with deis

Actually, I'd prefer we amend https://ci.deis.io/job/workflow-test-seed-job/ instead (and delete `recoverability_seed_job` if/when no longer needed.  Holler at me if you need help with this...

jobs/storage_test.groovy
jchauncey: Yeah it would be awesome if STORAGE_TYPE had a default but could be specified when running the job.

smothiki:
thanks changes incoming

smothiki:
@vdice @jchauncey  made changes accordingly PTAL

jobs/storage_test.groovy
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  steps {
+    shell '''
+      #!/usr/bin/env bash
+
+      set -eo pipefail
+
+      rand="${RANDOM}"
+      echo $rand
```
vdice: So now that the user can supply a different storage backend besides gcs, don't we need logic here to support, say, setting up s3/azure buckets?  If the script starts to get ungainly, we might consider moving it to `deis/charts` and then we can just call it here, like we do `./ci.sh`... could be something like `./setup_storage_prereqs.sh`...

vdice: did you still need the `echo $rand`?

smothiki: Apologies for not addressing your previous question. I had typed the answer and forgot to click comment which is in another tab. For the 1st question 

> If the script starts to get ungainly, we might consider moving it to deis/charts and then we can just call it here, like we do ./ci.sh

I wouldn't try my luck with azure. But the new change on the PR should be sufficient enough to configure storage backends to AWS ot GCS.
for 2nd one. 

> did you still need the echo $rand? 

printing rand to keep track of buckets created for a particular Job . let me know if there is another way to do this insteas of rand how about ${BUILD_NUMBER}

vdice: @smothiki ah ok.  printing `rand` is fine for this PR, but we can set up actually handing down `rand` as a parameter to downstream jobs if/when that is needed.

smothiki: Im printing rand to correlate which buckets correspond to a job run. let me know how to do that so that we can avoid rand

vdice: @smothiki I missed this; in the Jenkins Job DSL you have access to `${BUILD_NUMBER}` at any point, so you could easily append this value to the bucket names instead of using `rand` for identifying the buckets afterwards...

jobs/storage_test.groovy
```diff
@@ -0,0 +1,69 @@
...
+      if [ ${STORAGE_TYPE} == "gcs" ]; then
+        TYPE="GCS"
+      else
+        TYPE="AWS"
+        export S3_REGION="us-west-1"
+      fi
+      export ${TYPE}_REGISTRY_BUCKET="deis-registry-${rand}"
```
vdice: as mentioned above, we could eliminate the use of `rand` and just do something like this:

```
export ${TYPE}_REGISTRY_BUCKET="deis-registry-${BUILD_NUMBER}"
export ${TYPE}_BUILDER_BUCKET="deis-builder-${BUILD_NUMBER}"
export ${TYPE}_DATABASE_BUCKET="deis-database-${BUILD_NUMBER}"
```

I'd say we might also add `jenkins`, `${JOB_NAME}` and/or another such identifier to be explicitly clear it was created by ci...

PS here is a list of some of the env vars we get from the job DSL: https://github.com/jenkinsci/job-dsl-plugin/wiki/User-Power-Moves#access-the-jenkins-environment-variables

smothiki:
thanks @jchauncey

PR #86, opened 2016-05-20
feat(e2e-runner): Use e2e-runner image to execute all e2e runs

jchauncey:
e2e-runner will lease a k8s cluster and install a workflow chart of your choosing and then start the e2e tests

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -22,6 +22,8 @@ repos.each { Map repo ->
 TEST_JOB_ROOT_NAME = 'workflow-test'
 
 defaults = [
+  tmpPath: '/tmp/${JOB_NAME}/${BUILD_NUMBER}'
+  envFile: "${tmpPath}/env.properties"
   daysToKeep: 14,
   bumpverCommitCmd: 'git commit -a -m "chore(versions): ci bumped versions via ${BUILD_URL}" || true',
   testJob: [
@@ -46,3 +48,15 @@ defaults = [
     remoteName: 'deis',
   ],
 ]
+
+E2E_RUNNER_JOB = '''
+#!/usr/bin/env bash
+set -eo pipefail
+
+export WORKFLOW_CHART="workflow-${RELEASE}"
+export WORKFLOW_E2E_CHART="workflow-${RELEASE}-e2e"
+
+mkdir -p ${E2E_DIR_LOGS}
+env > ${E2E_DIR}/env.file
+docker run -u jenkins:jenkins --env-file=${E2E_DIR}/env.file -v ${E2E_DIR_LOGS}:/home/jenkins/logs:rw $E2E_RUNNER_IMAGE
+'''.stripIndent()
--- a/jobs/e2e-runner.groovy
+++ b/jobs/e2e-runner.groovy
@@ -1,0 +1,126 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+import utilities.StatusUpdater
+
+[ [type: 'master', branch: 'master'],
+  [type: 'pr', branch: '${sha1}'],
+].each { Map config ->
+  isMaster = config.type == 'master'
+  isPR = config.type == 'pr'
+  name = isMaster ? "e2e-runner" : "e2e-runner-${config.type}"
+  downstreamJobName = defaults.testJob[config.type]
+
+  job(name) {
+    description """
+      Watches the e2e-runner repo for a ${config.type} commit
+    """.stripIndent().trim()
+
+    scm {
+      git {
+        remote {
+          github("deis/e2e-runner")
+          credentials('597819a0-b0b9-4974-a79b-3a5c2322606d')
+          if (isPR) {
+            refspec('+refs/pull/*:refs/remotes/origin/pr/*')
+          }
+        }
+        branch(config.branch)
+      }
+    }
+
+    if (isPR) {
+      concurrentBuild()
+      throttleConcurrentBuilds {
+        maxPerNode(defaults.maxBuildsPerNode)
+        maxTotal(defaults.maxTotalConcurrentBuilds)
+      }
+    }
+
+    logRotator {
+      daysToKeep defaults.daysToKeep
+    }
+
+    if (isPR) { // set up GitHubPullRequest build trigger
+      triggers {
+        pullRequest {
+          admin('deis-admin')
+          cron('H/5 * * * *')
+          useGitHubHooks()
+          triggerPhrase('OK to test')
+          orgWhitelist(['deis'])
+          allowMembersOfWhitelistedOrgsAsAdmin()
+          // this plugin will update PR status no matter what,
+          // so until we fix this, here are our default messages:
+          extensions {
+            commitStatus {
+              context('ci/jenkins/pr')
+              triggeredStatus("Triggering e2e-runner build...")
+              startedStatus("Starting e2e-runner build...")
+              completedStatus('SUCCESS', "Merge with caution! Test job(s) may still be in progress...")
+              completedStatus('FAILURE', 'Build/deploy returned failure(s).')
+              completedStatus('ERROR', 'Something went wrong.')
+            }
+          }
+        }
+      }
+    }
+
+    parameters {
+      stringParam('DOCKER_USERNAME', 'deisbot', 'Docker Hub account name')
+      stringParam('DOCKER_EMAIL', 'dummy-address@deis.com', 'Docker Hub email address')
+      stringParam('QUAY_USERNAME', 'deisci+jenkins', 'Quay account name')
+      stringParam('QUAY_EMAIL', 'deisci+jenkins@deis.com', 'Quay email address')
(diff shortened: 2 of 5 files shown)


jchauncey:
Ok the last problem to really solve is how we build the e2e-runner.

jobs/workflow_test.groovy
```diff
@@ -139,13 +141,12 @@ import utilities.StatusUpdater
...
       shell """
         #!/usr/bin/env bash
-
         set -eo pipefail
-
-        ./ci.sh
-        if [ "\${COMMIT}" == "true" ]; then
-          ${defaults.bumpverCommitCmd}
-        fi
+        env > /tmp/\${JOB_NAME}/\${BUILD_NUMBER}/env.file
```
vdice: just a note: since we don't have to interpolate any groovy/DSL-defined variables (as we did before with `${defaults.bumpverCommitCmd}`, we could change this triple-double-quoted string to a triple-single-quoted string and then we wouldn't have to escape the `$`'s

jobs/workflow_test_release.groovy
```diff
@@ -66,16 +66,21 @@ job(name) {
     colorizeOutput 'xterm'
     credentialsBinding {
       string("GCLOUD_CREDENTIALS", "246d6550-569b-4925-8cda-e11a4f0d6803")
+      string("AUTH_TOKEN", "a62d7fe9-5b74-47e3-9aa5-2458ba32da52")
```
vdice, line 61: what's `AUTH_TOKEN`?  For authorizing the docker actions?  Any way we can name it a bit more descriptively?

jchauncey, line 61: no because thats what the k8s-claimer cli expects

jobs/e2e-runner.groovy
```diff
@@ -0,0 +1,105 @@
...
+
+import utilities.StatusUpdater
+
+[ [type: 'master', branch: 'master'],
+  [type: 'pr', branch: '${sha1}'],
+].each { Map config ->
+  isMaster = config.type == 'master'
+  isPR = config.type == 'pr'
+
+  name = isMaster ? "e2e-runner-master" : "e2e-runner-pr"
```
vdice: you could just say `name = "e2e-runner-${config.type}"` here

vdice: or actually I think our precedent was master job = `"e2e-runner"` and pr job = `"e2e-runner-pr"`... if thats the case and we want to stick w/ convention, we'd keep the same logic, minus the `"master"` in the master job name

jobs/workflow_test_release.groovy
```diff
@@ -66,16 +66,21 @@ job(name) {
...
   steps {
     shell """
       #!/usr/bin/env bash
-
       set -eo pipefail
-
-      WORKFLOW_CHART="workflow-\${RELEASE}" WORKFLOW_E2E_CHART="workflow-\${RELEASE}-e2e" ./ci.sh
+      export WORKFLOW_CHART="workflow-\${RELEASE}"
+      export WORKFLOW_E2E_CHART="workflow-\${RELEASE}-e2e"
+      env > /tmp/\${JOB_NAME}/\${BUILD_NUMBER}/env.file
```
vdice: since both `workflow-test(-pr)` and `workflow-test-release` jobs appear to run the same logic contained on these lines 79-83, wondering if we can extract it into either a common/re-usable string in `common.groovy` or a similarly accessible bash script that multiple jobs can reference?

jobs/workflow_test_release.groovy
```diff
@@ -66,16 +66,21 @@ job(name) {
...
+    shell '''
       #!/usr/bin/env bash
-
       set -eo pipefail
-
-      WORKFLOW_CHART="workflow-\${RELEASE}" WORKFLOW_E2E_CHART="workflow-\${RELEASE}-e2e" ./ci.sh
-    """.stripIndent().trim()
+      export WORKFLOW_CHART="workflow-${RELEASE}"
+      export WORKFLOW_E2E_CHART="workflow-${RELEASE}-e2e"
+      env > /tmp/${JOB_NAME}/${BUILD_NUMBER}/env.file
```
vdice: oops; recommenting as it got nuked on an outdated diff: since both workflow-test(-pr) and workflow-test-release jobs appear to run the same logic contained on these lines 79-83, wondering if we can extract it into either a common/re-usable string in common.groovy and/or a similarly accessible bash script that multiple jobs can reference?

jchauncey: k updated

common.groovy
```diff
@@ -40,3 +40,15 @@ defaults = [
     channel: '#testing',
   ],
 ]
+
+E2E_RUNNER_JOB = '''
```
vdice, line 52: nice -- but `workflow-test(-pr)` doesn't define `RELEASE` as far as I'm aware... so we could either define it as `dev` in that job or move the two exports to only the release test job...

jchauncey, line 52: fixed

bacongobbler:
looks like a rebase is in order

jobs/e2e-runner.groovy
```diff
@@ -0,0 +1,127 @@
...
+].each { Map config ->
+  isMaster = config.type == 'master'
+  isPR = config.type == 'pr'
+
+  name = isMaster ? "e2e-runner" : "e2e-runner-${config.type}"
+  downstreamJobName = defaults.testJob[config.type]
+
+  job(name) {
+    description """
+      Watches the workflow-e2e repo for a ${config.type} commit
```
vdice: s/workflow-e2e/e2e-runner

jobs/e2e-runner.groovy
```diff
@@ -0,0 +1,127 @@
...
+          useGitHubHooks()
+          triggerPhrase('OK to test')
+          orgWhitelist(['deis'])
+          allowMembersOfWhitelistedOrgsAsAdmin()
+          // this plugin will update PR status no matter what,
+          // so until we fix this, here are our default messages:
+          extensions {
+            commitStatus {
+              context('ci/jenkins/pr')
+              triggeredStatus("Triggering workflow-e2e/e2e-runner build...")
```
vdice: could just interpolate `name` here: `"Triggering ${name} build"` -- or even just use the Jenkins var (should be same) `"Triggering ${JOB_NAME} build"` and below/elsewhere

jobs/e2e-runner.groovy
```diff
@@ -0,0 +1,127 @@
...
+
+        mkdir -p "/tmp/${JOB_NAME}~/${BUILD_NUMBER}
+        eval $(make image)
+        # if triggered by pull request plugin, use ghprbActualCommit
+        export ACTUAL_COMMIT="\${ghprbActualCommit}"
+        # if manually triggered, use sha1
+        if [ -z "\${ghprbActualCommit}" ]; then
+        	export ACTUAL_COMMIT="\${sha1}"
+        fi
+        echo ACTUAL_COMMIT="\${ACTUAL_COMMIT}" >> /tmp/${JOB_NAME}/${BUILD_NUMBER}/env.properties
```
vdice: might consider a var in this groovy file to hold the `env.properties` filepath so it only needs updating in one place:

```
propertiesFilepath = "/tmp/${JOB_NAME}/${BUILD_NUMBER}/env.properties"
```

then can interpolate in shell script as usual: `${propertiesFilepath}` here and below in the `propertiesFile(propertiesFilepath)`

jobs/e2e-runner.groovy
jchauncey: not really a need its not saving us anything to do interpolation here since this job isnt being replicated in other places.

jobs/workflow_test_release.groovy
```diff
@@ -32,24 +22,28 @@ job(name) {
...
-   stringParam('HELM_REMOTE_NAME', defaults.helm["remoteName"], "Helm remote name")
-   stringParam('RELEASE', defaults.workflowRelease, "Release string for resolving workflow-[release](-e2e) charts")
+   stringParam('WORKFLOW_BRANCH', 'master', "The branch to use for installing the workflow chart.")
+   stringParam('WORKFLOW_E2E_BRANCH', 'master', "The branch to use for installing the workflow-e2e chart.")
+   stringParam('RELEASE', 'rc1', "Release string for resolving workflow-[release](-e2e) charts")
+   stringParam('HELM_REMOTE_REPO', 'https://github.com/deis/charts.git', "The remote repo to use for fetching charts.")
+   stringParam('E2E_RUNNER_IMAGE', 'quay.io/deisci/e2e-runner:latest', "The e2e-runner image")
+   stringParam('E2E_DIR', '$BUILD_NUMBER', "Directory for storing workspace files")
+   stringParam('E2E_DIR_LOGS', '$BUILD_NUMBER/logs', "Directory for storing logs. This directory is mounted into the e2e-runner container")
+>>>>>>> feat(e2e-runner): Use e2e-runner image to execute all e2e runs
```
vdice: looks like a merge conflict/rebase relic

PR #93, opened 2016-05-27
fix(workflow_test): Remove git push on success

jchauncey:


sgoings:
After the removal of https://github.com/deis/jenkins-jobs/pull/93/files#diff-fb7d1cdab123fee3519f7474673bc355L118 I'll LGTM. Per @vdice's statement:

> totes can remove `COMMIT` — weird naming parallels with `ACTUAL_COMMIT` but former was a boolean for, if master job, then we should commit/git push (when we did bumpver) while later holds a sha string

jchauncey:
@sgoings updated

PR #102, opened 2016-06-10
ref(website): jerbs for deis.com and workflow docs

slack:
- add pull-request runner for deis.com
- rename workflow -> workflow-docs

TODO:
- [x] after merge delete the old `workflow` job
- [x] after merge delete the old `deis_com` job
- [x] after merge delete ethe test seed job


slack:
Ok, @vdice consolidated the deis.com stuff into `deis_com.groovy` and ran a pr test.

Also renamed deis_com to deis_com_master which will trigger deploy_website as a downstream.

PR #104, opened 2016-06-13
feat(jobs/component_jobs.groovy): support testing sibling commits

### LINKED ISSUE #54: How to test 'paired' changes in separate repos
Our CI pipeline currently only supports running e2e against one specific commit in one repo.  However, our [e2e jenkins job](https://github.com/deis/jenkins-jobs/blob/master/jobs/workflow_test.groovy) in isolation actually does have the capacity to track and test multiple commits (see all `<COMPONENT>_SHA` env vars [here](https://github.com/deis/jenkins-jobs/blob/master/jobs/workflow_test.groovy#L102-L104)).

One angle might be to make use of the `ghprbLongDescription` value that we get in using the [GitHub Pull Request Builder Plugin](https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin).  As long as the committer supplied a standard message in the commit body, something along the lines of:

```
Requires slugrunner sha 80281d0
Requires builder sha 45ac6da
...
```

and so on, for the one or more required/'paired' changes, then our test job could feasibly be provided said sha's in the form of appropriate env vars from above.

vdice:
This introduces an additional shell script into every Workflow component job to check if the `${ghprbPullLongDescription}` value (essentially, the commit message body, provided by the [GitHub Pull Request Builder Plug-in](https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin)) contains any 'required' associated commits to include when running the downstream test job.

The format of said commit requiring proposed in this PR would be: `"[rR]equires repo#sha"`, e.g. 

```
requires builder#abc1234
Requires controller#ghi9123
```

TODO
- [x] PR related mention of functionality in https://github.com/deis/workflow/blob/master/src/contributing/submitting-a-pull-request.md
- [x] delete https://ci.deis.io/job/sibling-commit-tester-seed-job/

Ref https://github.com/deis/workflow/pull/321
Closes #54

### DIFF
--- a/bash/Dockerfile
+++ b/bash/Dockerfile
@@ -1,0 +1,23 @@
+FROM gcr.io/google_containers/ubuntu-slim:0.3
+
+ENV BATS_VERSION 0.4.0
+
+RUN addgroup --gid 999 batsman
+RUN adduser --system \
+	--shell /bin/bash \
+	--disabled-password \
+	--gid 999 \
+	--uid 999 \
+	batsman
+
+RUN apt-get update -q \
+	&& apt-get install -y -q --no-install-recommends bash make curl ca-certificates \
+	&& curl -o "/tmp/v${BATS_VERSION}.tar.gz" -L \
+		"https://github.com/sstephenson/bats/archive/v${BATS_VERSION}.tar.gz" \
+	&& tar -x -z -f "/tmp/v${BATS_VERSION}.tar.gz" -C /tmp/ \
+	&& bash "/tmp/bats-${BATS_VERSION}/install.sh" /usr/local \
+	&& rm -rf /tmp/*
+
+USER batsman
+
+ENTRYPOINT ["/usr/local/bin/bats"]
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -118,18 +118,27 @@ repos.each { Map repo ->
           echo ACTUAL_COMMIT="\${ACTUAL_COMMIT}" > \${WORKSPACE}/env.properties
         """.stripIndent().trim()
 
+        shell new File("${WORKSPACE}/bash/scripts/commit_description_parser.sh").text
+
         // do not run e2e tests for workflow-manager at this time
         if (repo.name != 'workflow-manager') {
-          downstreamParameterized {
-            trigger(downstreamJobName) {
-              parameters {
-                propertiesFile('${WORKSPACE}/env.properties')
-                predefinedProps([
-                  "${repo.commitEnvVar}": '${GIT_COMMIT}',
-                  'UPSTREAM_BUILD_URL': '${BUILD_URL}',
-                  'UPSTREAM_SLACK_CHANNEL': "${repo.slackChannel}",
-                  'COMPONENT_REPO': "${repo.name}",
-                ])
+          conditionalSteps {
+            condition {
+              shell new File("${WORKSPACE}/bash/scripts/skip_e2e_check.sh").text
+            }
+            steps {
+              downstreamParameterized {
+                trigger(downstreamJobName) {
+                  parameters {
+                    propertiesFile('${WORKSPACE}/env.properties')
+                    predefinedProps([
+                      "${repo.commitEnvVar}": '${GIT_COMMIT}',
+                      'UPSTREAM_BUILD_URL': '${BUILD_URL}',
+                      'UPSTREAM_SLACK_CHANNEL': "${repo.slackChannel}",
+                      'COMPONENT_REPO': "${repo.name}",
+                    ])
+                  }
+                }
               }
             }
           }
--- a/bash/Makefile
+++ b/bash/Makefile
@@ -1,0 +1,12 @@
+TEST_CMD := --tap tests
+
+test:
+	bats ${TEST_CMD}
+
+docker-build:
+	docker build -t bats .
+
+docker-test: docker-build
+	docker run -v ${CURDIR}:/bash -w /bash bats ${TEST_CMD}
+
+.PHONY: test docker-test docker-build
--- a/bash/scripts/commit_description_parser.sh
+++ b/bash/scripts/commit_description_parser.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+main() {
+  envPropsFilepath="/dev/null"
+  if [ -n "${JENKINS_HOME}" ] && [ "${JOB_BASE_NAME}" != "jenkins-jobs" ]; then
+    envPropsFilepath="${WORKSPACE}/env.properties"
+  fi
+  parse-commit-description "${ghprbPullLongDescription}" >> "${envPropsFilepath}"
+}
+
+# parse-commit-description parses a commit description for any required sibling
+# to pass along to downstream job(s)...
+parse-commit-description() {
+  description="${1}"
+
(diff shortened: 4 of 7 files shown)


vdice:
Question for reviewers:  Right now we look for `req <repo_name>#<commit_sha>` in the commit body, as in “requires repo_name commit_sha", but should we just use `ref(s)` instead?  

(We use `ref` already to refer to associated documentation or even e2e PRs... both of which technically could be used to derive the Docker image to reference for the downstream test job, assuming they exist in the `deisci` org at test time...)

arschles:
@vdice IMO we should use `requires`, not `ref`

vdice:
@arschles should we use the long form of `requires` over `req` as well?  

Instead of `req slugbuilder#abc1234` we would expect `requires slugbuilder#abc1234`.

arschles:
I'd rather use the long form. More explicit is better IMO, even if it requires a few more keystrokes

vdice:
Also, thinking for documenting how to use this functionality in https://github.com/deis/workflow/blob/master/src/contributing/submitting-a-pull-request.md.  Does this seem like the proper spot?

arschles:
yup, that seems like the best spot to start

vdice:
@arschles now updated to look for `[Rr]equires` in https://github.com/deis/jenkins-jobs/pull/104/commits/ba27f47645da98dde26d2d7cc576b7e982db5a7b

jobs/component_jobs.groovy
```diff
@@ -118,6 +118,30 @@ repos.each { Map repo ->
...
         """.stripIndent().trim()
 
+        shell '''
+          #!/usr/bin/env bash
+          set -eo pipefail
+
+          # Commit description parser for passing along required sibling commits...
+
+          # Looks specifically for matches of '[rR]eq(s) <repo>#<sha>',
+          # e.g., "req builder#abc1234, Reqs router#def5678"
```
arschles: can you update these commands to reference matches of `[Rr]equires`?

vdice: Thanks @arschles; now updated.

vdice:
ping @jchauncey and @mboersma.  Logic has been updated per conversation in related docs issue https://github.com/deis/workflow/pull/321.  Please revisit when convenient.  Thanks!

helgi:
It can't be a PR number instead of a sha?

vdice:
@helgi I think we could but I would rather iterate on what we have currently if that's okay.  I'd like to see how this change works from a CI angle to see what issues may arise.  

I can definitely create a refactor ticket to allow specifying the PR number instead of commit, which would probably look like a curl command to the following API [endpoint](https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request) (we do have a similar example curling the GH API in the [status updater](https://github.com/deis/jenkins-jobs/blob/master/jobs/utilities/StatusUpdater.groovy))

Edit: ticket created in https://github.com/deis/jenkins-jobs/issues/111

jobs/component_jobs.groovy
```diff
@@ -118,18 +118,60 @@ repos.each { Map repo ->
...
 
+        shell '''
+          #!/usr/bin/env bash
+          set -eo pipefail
+
+          # Commit description parser for passing along required sibling commits...
+
+          # Looks specifically for matches of '[rR]equires <repo>#<sha>',
+          # e.g., "requires builder#abc1234, Requires router#def5678"
+          reqs=`echo "${ghprbPullLongDescription}" | grep -o "[Rr]equires [-a-z]*#[a-z0-9]*" | grep -o "[-a-z]*#[a-z0-9]*"` || true
```
sgoings: Why the `|| true`?

sgoings:
Some Q's
1. What ensures that the SHAs referenced by the `requires` actually exist in Quay by the time this job runs?
2. This seems dang hard to test without uploading to Jenkins. Have you thought about ways you might be able to test this script in a standalone fashion? I'm not necessarily 🚫  the merge of this PR, but I'd want to see this be anti-brittle... just in case I stumbled upon this repository late one night with nothing better to do but make this more Groovy. 😉

jobs/component_jobs.groovy
vdice: Perhaps there is a `grep` flag that would be better to use, but currently, if this `grep` command does not find a match, it exits non-zero.  Since we find it acceptable to not find a match (which constitutes the normal case for PRs not requiring others) and I've got `set -eo pipefail` above, we don't want this non-zero exit code to cause the script itself to exit non-zero.

vdice:
@sgoings:

> What ensures that the SHA referenced by the requires actually exists in Quay by the time this job runs?

Nothing currently.  Therefore, the test job will fail if one or more SHAs referenced to not exist in quay.

> This seems dang hard to test without uploading to Jenkins. Have you thought about ways you might be able to test this script in a standalone fashion? I'm not necessarily 🚫 the merge of this PR, but I'd want to see this be anti-brittle... just in case I stumbled upon this repository late one night with nothing better to do but make this more Groovy.

Though I had tested it via one-off jobs in Jenkins (but only insofar as to verify the shas were being handed to the downstream test job), I def agree we are lacking a lower-level test framework for the shell logic we currently have in this repo.  I have some experience with the bats testing framework which I could bring in for testing the shell logic added here in this PR or could move this task into a ticket to pull in immediately afterwards -- preference?

sgoings:
> Nothing currently. Therefore, the test job will fail if one or more SHAs referenced to not exist in quay.

This is definitely going to cause some pain, but at least @helgi will be happier?

vdice:
> This is definitely going to cause some pain

@sgoings yea, you are right... we decided the initial workaround for the user would look like the docs PR'd here: https://github.com/deis/workflow/pull/321.  Take a gander and let me know if another approach might be suitable...

vdice:
@sgoings bash logic added as part of this PR is now covered under test.  Please revisit when convenient!

sgoings:
Running the bats tests locally is 👍 👍 👍 

```
docker run -v /Users/sethgoings/code/jenkins-jobs/bash:/bash -w /bash bats --tap tests
1..9
ok 1 parse-commit-description : no description given
ok 2 parse-commit-description : description with no requirements
ok 3 parse-commit-description : description with requirements
ok 4 parse-commit-description : description with ill-formated requirements
ok 5 main : if on jenkins in jenkins-jobs workspace
ok 6 main : if on jenkins and not in jenkins-jobs workspace
ok 7 check-skip-e2e : no description given
ok 8 check-skip-e2e : description with no 'skip e2e'
ok 9 check-skip-e2e : description with 'skip e2e
```

bash/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM gcr.io/google_containers/ubuntu-slim:0.3
+
+ENV BATS_VERSION 0.4.0
+
+RUN addgroup --gid 999 batsman
```
mboersma, line 5: "batsman"--well played sir, like a cricketeer.

bash/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN adduser --system \
+	--shell /bin/bash \
+	--disabled-password \
+  --gid 999 \
```
mboersma: Could you fix the inconsistent indentation here?

vdice: Thanks @mboersma, fixed.

PR #107, opened 2016-06-14
chore(workflow-test): Use e2e-runner:canary

jchauncey:


arschles:
Does this require https://github.com/deis/e2e-runner/pull/18?

PR #108, opened 2016-06-14
fix(e2e_runner.groovy): always push both immutable and mutable e2e-runner images

arschles:
On pushes to `master`.

Addresses test failures like https://ci.deis.io/job/workflow-test/624/testReport

### DIFF
--- a/jobs/e2e_runner.groovy
+++ b/jobs/e2e_runner.groovy
@@ -8,7 +8,7 @@ import utilities.StatusUpdater
   isMaster = config.type == 'master'
   isPR = config.type == 'pr'
   name = isMaster ? "e2e-runner" : "e2e-runner-${config.type}"
-  dockerPush = isPR ? 'docker-immutable-push' : 'docker-mutable-push'
+  dockerPush = 'docker-push'
   downstreamJobName = defaults.testJob[config.type]
 
   job(name) {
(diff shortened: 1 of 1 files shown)


jobs/e2e_runner.groovy
```diff
@@ -8,7 +8,7 @@ import utilities.StatusUpdater
   isMaster = config.type == 'master'
   isPR = config.type == 'pr'
   name = isMaster ? "e2e-runner" : "e2e-runner-${config.type}"
-  dockerPush = isPR ? 'docker-immutable-push' : 'docker-mutable-push'
+  dockerPush = 'docker-push'
```
vdice: I think we still want to _only_ push the immutable tag for a PR (so `canary` doesn't get a potentially flawed PR change), but we do want to push both immutable and mutable on a master commit (both the immutable tag for the actual merge commit and the updated `canary` tag)... something like:

```
dockerPush = isPR ? 'docker-immutable-push' : 'docker-push'
```

(This is what we do in our existing `component_job.groovy` https://github.com/deis/jenkins-jobs/blob/master/jobs/component_jobs.groovy#L98)

arschles: ahhh, I see what you mean, and :+1: on pushing only the immutable. change incoming

arschles: fixed in https://github.com/deis/jenkins-jobs/pull/108/commits/e21a13b7bde8f7c5d3bf43c2e1180f2ae8880836

PR #113, opened 2016-06-21
feat(jenkins_jobs_pr.groovy): add jenkins-jobs-pr job

vdice:
To run tests against a PR commit

Depends on https://github.com/deis/jenkins-jobs/pull/104

### DIFF
--- a/jobs/jenkins_jobs_pr.groovy
+++ b/jobs/jenkins_jobs_pr.groovy
@@ -1,0 +1,67 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+name = 'jenkins-jobs-pr'
+
+job(name) {
+  description """
+    <ol>
+      <li>Watches the <a href="https://github.com/deis/jenkins-jobs">jenkins-jobs</a> repo for a PR commit</li>
+      <li>and runs tests against changes</li>
+    </ol>
+  """.stripIndent().trim()
+
+  scm {
+    git {
+      remote {
+        github("deis/jenkins-jobs")
+        credentials('597819a0-b0b9-4974-a79b-3a5c2322606d')
+      }
+      refspec('+refs/pull/*:refs/remotes/origin/pr/*')
+      branch('${sha1}')
+    }
+  }
+
+  logRotator {
+    daysToKeep defaults.daysToKeep
+  }
+
+  triggers {
+    githubPush()
+
+    pullRequest {
+      admin('deis-admin')
+      cron('H/5 * * * *')
+      useGitHubHooks()
+      triggerPhrase('OK to test')
+      orgWhitelist(['deis'])
+      allowMembersOfWhitelistedOrgsAsAdmin()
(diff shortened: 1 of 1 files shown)


jobs/jenkins_jobs_pr.groovy
```diff
@@ -0,0 +1,65 @@
...
+    colorizeOutput 'xterm'
+  }
+
+  steps {
+    shell '''
+      #!/usr/bin/env bash
+
+      set -eo pipefail
+
+      make docker-test
```
sgoings, line 64: There's no Makefile at the root of the repository... only in the `bash` directory. Do you want to move the Makefile up a dir?

vdice, line 64: @sgoings ah good catch, thank you!  I've updated change to `cd` into the `bash` subdir for now.  Was thinking to keep the `Makefile` there for now as all targets relate to that particular subdir.

jobs/jenkins_jobs_pr.groovy
```diff
@@ -0,0 +1,67 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+name = 'jenkins-jobs-pr'
+
+job(name) {
+  description """
+    <ol>
+      <li>Watches the <a href="https://github.com/deis/jenkins-jobs">jenkins-jobs</a> repo for a PR commit</li>
+      <li>And runs tests against changes</li>
```
mboersma: Should `And` be lower-case? Extreme nitpick mode activated.

jobs/jenkins_jobs_pr.groovy
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  triggers {
+    githubPush()
+
+    pullRequest {
+      admin('deis-admin')
+      cron('H/5 * * * *')
+      useGitHubHooks()
+      triggerPhrase('OK to test')
```
mboersma, line 35: I've gotten used to commenting "Jenkins, test this please" to kick off a job. Does this replace that, or are they different trigger phrases?

vdice, line 35: @mboersma there must be other default trigger phrases, perhaps built into the plugin (though an initial cursory search doesn't seem to show [any](https://github.com/janinko/ghprb/search?utf8=%E2%9C%93&q=triggerPhrase)... fwiw we've used `OK to test` in our jobs since we started converting them to DSL.

jobs/jenkins_jobs_pr.groovy
vdice: Extreme nitpick annihilated! :)

jobs/jenkins_jobs_pr.groovy
mboersma, line 35: >  there must be other default trigger phrases

I think you're right--I know "test this please" has worked on existing jenkins-jobs. :+1:

PR #115, opened 2016-06-21
fix(common.groovy): pull E2E_RUNNER_IMAGE

vdice:
...since oftentimes the tag will be 'canary' and we don't want
to use whatever might be cached on the Jenkins host


jchauncey:
thats a really good catch!

PR #123, opened 2016-06-23
feat(e2e_runner.groovy): add 'make docker-test' step

vdice:
Depends on https://github.com/deis/e2e-runner/pull/24


vdice:
(as noted in description, ci will fail until https://github.com/deis/e2e-runner/pull/24 gets merged)

PR #124, opened 2016-06-24
feat(workflow_test*.groovy): add WORKFLOW_CLI_SHA param

vdice:
So a specific version can be used; also fixed funky indentation

cc @joshua-anderson

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -27,6 +27,7 @@ set -eo pipefail
 
 export WORKFLOW_CHART="workflow-${RELEASE}"
 export WORKFLOW_E2E_CHART="workflow-${RELEASE}-e2e"
+export CLI_VERSION="${WORKFLOW_CLI_SHA:0:7}"
 
 mkdir -p ${E2E_DIR_LOGS}
 env > ${E2E_DIR}/env.file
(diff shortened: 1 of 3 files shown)


vdice:
~~CI will fail until https://github.com/deis/jenkins-jobs/pull/116 is merged~~

Joshua-Anderson:
LGTM, though I'm not familiar with the Jenkins settings parts.

common.groovy
```diff
@@ -27,6 +27,7 @@ set -eo pipefail
 
 export WORKFLOW_CHART="workflow-${RELEASE}"
 export WORKFLOW_E2E_CHART="workflow-${RELEASE}-e2e"
+export CLI_VERSION="${WORKFLOW_CLI_SHA:0:7}"
```
Joshua-Anderson, line 30: You actually only want the first `6` characters of the SHA, which is the form the binary uses to upload: `https://storage.googleapis.com/workflow-cli/e233d34/deis-e233d34-linux-amd64`

Joshua-Anderson, line 30: And I can't count at all... don't mind me 😄

PR #126, opened 2016-06-27
chore(jobs/workflow_test.groovy): explicitly define upstream params

vdice:
Pre Jenkins 2.3, one could declare `predefinedProps` in an upstream job
and have those properties automatically be created in a downstream job.

Post-2.3, due to [security considerations](https://issues.jenkins-ci.org/browse/JENKINS-34871)
it appears any/all downstream jobs need to explicitly declare any properties/values
that it may be passed in a given pipeline.

This particular fix involves getting the correct values for `UPSTREAM_BUILD_URL` and `UPSTREAM_SLACK_CHANNEL`.  It also fixes incorrect indentation/spacing.

cc @mboersma

### DIFF
--- a/jobs/workflow_test.groovy
+++ b/jobs/workflow_test.groovy
@@ -90,6 +90,8 @@ import utilities.StatusUpdater
         stringParam(repo.commitEnvVar, '', "${repo.name} commit SHA")
       }
       stringParam('WORKFLOW_CLI_SHA', '', "workflow-cli commit SHA")
+      stringParam('UPSTREAM_BUILD_URL', '', "Upstream build url")
+      stringParam('UPSTREAM_SLACK_CHANNEL', '', "Upstream Slack channel")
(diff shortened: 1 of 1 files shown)


mboersma:
@vdice you're way ahead of me. Just noticed these warnings in the Jenkins logs and was going to ping you about it:

```
Skipped parameter `UPSTREAM_BUILD_URL` as it is undefined on `workflow-test-pr`. Set `-Dhudson.model.ParametersAction.keepUndefinedParameters`=true to allow undefined parameters to be injected as environment variables or `-Dhudson.model.ParametersAction.safeParameters=[comma-separated list]` to whitelist specific parameter names, even though it represents a security breach
```

vdice:
@mboersma saw your mention this morning (it had evaded my notice) so indeed did some research and it looks like this had been broken since our Jenkins upgrade to >= 2.3

PR #129, opened 2016-06-30
feat(component_release.groovy): add <component>-release job

### LINKED ISSUE #110: create component release job
We have the need for a `component-release-job` that would track git push events to a `release*` branch on any/all of our Workflow component [repos](https://github.com/deis/jenkins-jobs/blob/master/common.groovy#L1-L17).

Use case: If someone creates a release branch, we want a Docker image created and pushed from that release branch. 

Further, if they modify that branch (i.e. from fixes, cherry-picking or rebase), this job should fire again and push a new image, ready for use in a release chart.
#### Implementation details:

As of this writing, we are leaning to one `.groovy` file set up to track all component repos, rather than adding separate `-release` jobs for each component via additional configuration in the existing `component_jobs.groovy` [file](https://github.com/deis/jenkins-jobs/blob/master/jobs/component_jobs.groovy).

vdice:
For all Workflow component repos, to watch for a commit on a given release branch,
specified via the RELEASE parameter.

Live example built from DSL on this branch: https://ci.deis.io/job/dockerbuilder-release/1/console

TODO
- [x] mark this as completed in https://github.com/deis/deisrel/issues/100 if/when merged
- [x] delete manually created `component-seed-job`

Closes #110 
Ref https://github.com/deis/workflow/pull/362

### DIFF
--- a/bash/scripts/check_release_branch.sh
+++ b/bash/scripts/check_release_branch.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+main() {
+  check-release-branch
+}
+
+check-release-branch() {
+  if [ "${GIT_BRANCH}" != "origin/release-${RELEASE}" ]; then
+    echo "GIT_BRANCH (${GIT_BRANCH}) is not 'origin/release-${RELEASE}'; exiting build."
+    exit 0
+  fi
+  echo "Proceeding with build/deploy from 'origin/release-${RELEASE}'..."
+}
--- a/bash/tests/check_release_branch_test.bats
+++ b/bash/tests/check_release_branch_test.bats
@@ -1,0 +1,23 @@
+#!/usr/bin/env bats
+
+setup() {
+  . "${BATS_TEST_DIRNAME}/../scripts/check_release_branch.sh"
+}
+
+@test "check-release-branch : is non-release branch" {
+  export RELEASE='foo'
+  export GIT_BRANCH='origin/master'
+  run check-release-branch
+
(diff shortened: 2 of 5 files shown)


vdice:
the only weirdness is w/ this DSL _any_ commit to the repo (say, master commit) triggers this job (though it does checkout the release branch and just build/deploy from the HEAD commit), for instance here is an example: https://ci.deis.io/job/controller-release/1/console

Could try adding `refspec('+refs/heads/release-${RELEASE}:refs/remotes/origin/release-${RELEASE}')` to the `git` block... not sure if that would change the behavior or not

Edit: it does not.  a githubPush event to the repo, regardless of branch on `origin`, will kick this off -- and though its inefficient and wastes a cycle on a node, it only _re-deploys_ from existing `HEAD` commit on the release branch, which I don't believe should be an issue unless I'm missing something...

vdice:
@krancour updated with your good idea of just checking if `GIT_BRANCH` matches `release-${RELEASE}`.  If yes, proceed; if no, simply exit 0.

vdice:
@krancour I also updated the `RELEASE` value to the next anticipated release, `v2.2.0`.  While no `release-v2.2.0` branch exists on any of the components, once they are created, these jobs will "just work".

Please review when convenient.

bacongobbler:
should we add some documentation in the release docs that explains what this release job will do? :)

vdice:
@bacongobbler definitely -- good call.  Will create the PR and ref it here asap.

vdice:
@bacongobbler docs PR added: https://github.com/deis/workflow/pull/362

PR #130, opened 2016-07-06
feat(misc.): add canonical WORKFLOW_RELEASE version

vdice:
So that update for new release need only happen in one place.

TODO:
- [x] update `component_release.groovy` to be added in https://github.com/deis/jenkins-jobs/pull/129

Ref https://github.com/deis/workflow/pull/363

### DIFF
--- a/jobs/workflow_test_release.groovy
+++ b/jobs/workflow_test_release.groovy
@@ -9,6 +9,16 @@ job(name) {
     <p>Runs a given workflow-[RELEASE]-e2e tests chart against a workflow-[RELEASE] chart candidate using e2e-runner</p>
   """.stripIndent().trim()
 
+  scm {
+    git {
+      remote {
+        github("deis/charts")
+        credentials('597819a0-b0b9-4974-a79b-3a5c2322606d')
+      }
+      branch('release-${RELEASE}')
+    }
+  }
+
   logRotator {
     daysToKeep defaults.daysToKeep
   }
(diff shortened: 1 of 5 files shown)


jobs/workflow_test_release.groovy
```diff
@@ -40,17 +50,17 @@ job(name) {
...
+    stringParam('RELEASE', defaults.workflow.release, "Release string for resolving workflow-[release](-e2e) charts")
+    stringParam('HELM_REMOTE_REPO', defaults.helm["remoteRepo"], "The remote repo to use for fetching charts.")
     stringParam('E2E_RUNNER_IMAGE', 'quay.io/deisci/e2e-runner:canary', "The e2e-runner image")
     stringParam('E2E_DIR', '/home/jenkins/workspace/$JOB_NAME/$BUILD_NUMBER', "Directory for storing workspace files")
     stringParam('E2E_DIR_LOGS', '${E2E_DIR}/logs', "Directory for storing logs. This directory is mounted into the e2e-runner container")
   }
 
   triggers {
-    cron('@daily')
+    githubPush()
```
mboersma, line 63: An unrelated change here. No need to split the commit, but could you explain this briefly @vdice? I assume daily was the wrong trigger.

vdice, line 63: Thanks for keeping me honest, @mboersma !  Indeed, strictly speaking, this should ideally have been a separate change.  However, to explain my reasoning:  

I wanted the release test job to trigger off of push events to release branches, rather than arbitrarily running them daily.  The real value of this job is during release time and I don't think there is value in running them arbitrarily during inter-release iterations -- though definitely receptive to feedback on this assumption.

mboersma, line 63: > The real value of this job is during release time and I don't think there is value in running them arbitrarily

I agree with that. Thanks for the explanation.

arschles:
LGTM, pending the unchecked TODO in the PR description

PR #131, opened 2016-07-06
chore(logger): Add jobs for redis logger component

krancour:


vdice:
@krancour LGTM assuming redis is considered a 'standard' Workflow component like the others.  Adding it to the list will generate the `redis` and `redis-pr` jobs, which must supply the standard containerized build/deploy `make` targets.  Also, both jobs will be wired by default to kick off the downstream e2e job -- just want to confirm that's what we are looking for here.

krancour:
@vdice -- on a trial basis, yes. We could end up reversing course on this, but for the moment, that was the intention.

Issue #111, opened 2016-06-20
add ability to specify PR number for commit dependencies


vdice:
https://github.com/deis/jenkins-jobs/pull/104 adds the ability to require other commits for the inclusion in the build/test pipeline.

It would be preferable if instead of having to supply the commit sha in `Requires <repo_name>#<commit_sha>` one could just reference the PR number: `Requires <repo_name>#<PR_number>` as is the usual protocol when referencing another issue/PR in GitHub.

The api call would probably be [this one](https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request) and see [here](https://github.com/deis/jenkins-jobs/blob/master/jobs/utilities/StatusUpdater.groovy) for an example of curl-ing the commit status GH API.

vdice:
pulling this in to fill in some 'backpocket' time in v2.2

PR #132, opened 2016-07-07
feat(commit_description_parser): take pr number instead

vdice:
And derive latest commit sha from pr number given
via appropriate GitHub API req.

Ref https://github.com/deis/workflow/pull/366
Closes https://github.com/deis/jenkins-jobs/issues/111

### DIFF
--- a/bash/scripts/commit_description_parser.sh
+++ b/bash/scripts/commit_description_parser.sh
@@ -10,25 +10,48 @@ main() {
 }
 
 # parse-commit-description parses a commit description for any required sibling
-# to pass along to downstream job(s)...
+# commits to pass along to downstream job(s)...
 parse-commit-description() {
   description="${1}"
 
-  # Looks specifically for matches of '[rR]equires <repo>#<sha>',
-  # e.g., "requires builder#abc1234, Requires router#def5678"
-  reqs=$(echo "${description}" | grep -o "[Rr]equires [-a-z]*#[a-z0-9]*" | grep -o "[-a-z]*#[a-z0-9]*") || true
+  # Looks specifically for matches of '[rR]equires <repo>#<prNumber>',
+  # e.g., "requires builder#123, Requires router#567"
+  reqs=$(echo "${description}" | grep -o "[Rr]equires [-a-z]*#[0-9]\{1,9\}" | grep -o "[-a-z]*#[0-9]\{1,9\}") || true
 
-  # split on whitespace into array of '<repo>#<sha>' values
+  # split on whitespace into array of '<repo>#<prNumber>' values
   reqsArray=(${reqs// / })
 
-  # for each '<repo>#<sha>' value in array
+  # for each '<repo>#<prNumber>' value in array
+  local i
   for i in "${reqsArray[@]}"; do
     # split on '#'
-    repoShaArray=(${i//#/ })
-    # echo '<uppercased repo>_SHA'=<sha> (with hyphens converted to underscores)
-    repoEnvVar=$(echo "${repoShaArray[0]//-/_}" | awk '{print toupper($0)}')_SHA
-    echo "${repoEnvVar}"="${repoShaArray[1]}"
+    repoPRNumberArray=(${i//#/ })
+    repoName="${repoPRNumberArray[0]}"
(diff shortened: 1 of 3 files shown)


bash/scripts/commit_description_parser.sh
```diff
@@ -10,25 +10,47 @@ main() {
...
   done
 }
 
+get-most-recent-pr-commit() {
+  repoName="${1}"
+  prNumber="${2}"
+
+  curl \
+  -sSL \
+  --user deis-admin:"\${GITHUB_ACCESS_TOKEN}" \
```
sgoings, line 52: Might want to move `deis-admin` to a variable.

bash/scripts/commit_description_parser.sh
```diff
@@ -10,25 +10,47 @@ main() {
...
+  "https://api.github.com/repos/deis/${repoName}/pulls/${prNumber}/commits" \
+  | docker run -i --rm kamermans/jq '.[].sha' | tail -n1
```
sgoings, line 54: This seems like a bit of overhead just to run jq. Not a huge deal for now... but might we have better options for running jq quickly?

sgoings:
Wooooo eeee. That bash test is advanced! 💯 

Tested manually: 👍

bash/scripts/commit_description_parser.sh
```diff
@@ -10,25 +10,47 @@ main() {
...
-# to pass along to downstream job(s)...
+# commits to pass along to downstream job(s)...
 parse-commit-description() {
   description="${1}"
 
-  # Looks specifically for matches of '[rR]equires <repo>#<sha>',
-  # e.g., "requires builder#abc1234, Requires router#def5678"
-  reqs=$(echo "${description}" | grep -o "[Rr]equires [-a-z]*#[a-z0-9]*" | grep -o "[-a-z]*#[a-z0-9]*") || true
+  # Looks specifically for matches of '[rR]equires <repo>#<prNumber>',
+  # e.g., "requires builder#123, Requires router#567"
```
sgoings, line 18: What happened to `4`? 😢

bash/scripts/commit_description_parser.sh
```diff
@@ -10,25 +10,47 @@ main() {
...
+  reqs=$(echo "${description}" | grep -o "[Rr]equires [-a-z]*#[0-9]*" | grep -o "[-a-z]*#[0-9]*") || true
```
sgoings: Slight regex edge case:

Would pass on:

`Requires controller#`

bash/scripts/commit_description_parser.sh
```diff
@@ -10,25 +10,47 @@ main() {
...
 
-  # split on whitespace into array of '<repo>#<sha>' values
+  # split on whitespace into array of '<repo>#<prNumber>' values
   reqsArray=(${reqs// / })
 
-  # for each '<repo>#<sha>' value in array
+  # for each '<repo>#<prNumber>' value in array
   for i in "${reqsArray[@]}"; do
```
sgoings, line 26: Above this you should reset `i`:

```
local i
```

If you don't, `i` can contain something else... and you'd process it below.

bash/scripts/commit_description_parser.sh
vdice, line 54: yea agreed, a bit 'heavy-handed' (though image only ~12MB).  Went this route b/c I knew ('linux') Jenkins slaves have docker, but not jq (yet).  Shall I create an issue to add `jq` to the Jenkins slaves so we can run directly -- and refactor this dependency out?

sgoings:
LGTM-ing due to mostly nits. Fixup as you please.

bash/scripts/commit_description_parser.sh
arschles, line 54: I'm sure you anticipated this comment from me... it seems like this suite of scripts is now complex enough to warrant building a Go program to replace (at least part of) it. Thoughts on doing that eventually?

vdice, line 54: @arschles mos def.  It's been on my mind as well.  We should brainstorm soon and generate a meta ticket w/ outline of transition.

PR #136, opened 2016-07-08
feat(milestone-watcher): add cron job to watch deis milestone

### LINKED ISSUE #133: Milestone watcher job
This proposes a solution to eliminating the manual [Step #2](https://github.com/deis/workflow/blob/master/src/roadmap/release-checklist.md#step-2-update-jenkins-jobs) from the [release-checklist.md](https://github.com/deis/workflow/blob/master/src/roadmap/release-checklist.md).  It currently involves cloning this repo, updating a value and pushing commit to master.

Instead, we can create a chron (daily?) job that can make the appropriate [GH api call](https://developer.github.com/v3/issues/milestones/#list-milestones-for-a-repository) and determine the next immediate 'open' milestone.  Once this value is acquired, the job would determine if it is different from the current [value](https://github.com/deis/jenkins-jobs/blob/master/common.groovy#L23).  If so, it would update via a commit to master.  If no, finish the build with a no-op.

vdice:
and update local `workflow.version` when necessary

Closes #133


bash/scripts/milestone_watcher.sh
```diff
@@ -0,0 +1,47 @@
...
+  "https://api.github.com/repos/deis/${repoName}/milestones" \
+  | docker run -i --rm kamermans/jq '.[].title' | head -n1
+}
+
+bump-workflow-version() {
+  version="${1}"
+
+  echo "${version}" > "$(get-workflow-version-filepath)"
+  git diff
+  # TODO: enable once this job gets a test flight...
```
vdice, line 35: test flight job: https://ci.deis.io/job/milestone-watcher/6/console

bash/scripts/milestone_watcher.sh
```diff
@@ -0,0 +1,47 @@
...
+  fi
+}
+
+retrieve-github-milestone() {
+  repoName="${1}"
+
+  curl \
+  -sSL \
+  "https://api.github.com/repos/deis/${repoName}/milestones" \
+  | docker run -i --rm kamermans/jq '.[].title' | head -n1
```
vdice, line 27: Note: this will pull the milestone verbatim from GitHub.  For example, the current release/milestone is precisely `v2.2` (whereas we had been in the habit of using the full-length semver for our releases, in this case, `v2.2.0`.)  We need to agree on what version will be the standard -- it will be easiest if the GitHub milestone is considered the exact string representation of the next release.

vdice:
@arschles I had a brainstorm around this.  I think the jenkins job itself should still be a similar cron job, but it might make more sense to put the actual GH api call/logic into https://github.com/deis/deisrel.   If in deisrel, those in charge of a particular release could then defer to the canonical GitHub milestone to set their `WORKFLOW_RELEASE(_SHORT)` values; something like:

`deisrel get-milestone` which would return the next open milestone from GH (`v2.2` at time of writing) and perhaps even echo the necessary export statements that the releaser(s) can invoke to set up their environment properly:

```
export WORKFLOW_RELEASE=<returned milestone>
export WORKFLOW_RELEASE_SHORT=<short version of returned milestone, if applicable>
```

This jenkins job would then just download the latest `deisrel` binary and run `deisrel get-milestone` on a cron basis, updating this repo's local version of the current Workflow release when they have diverged.

vdice:
Closing for now, can re-open if deemed helpful at a later date.

PR #141, opened 2016-07-12
fix(github creds): don't escape $ in bash env

vdice:
and refactor to only change GitHub token (via credentialsID) in one location.

### DIFF
--- a/bash/scripts/commit_description_parser.sh
+++ b/bash/scripts/commit_description_parser.sh
@@ -49,7 +49,7 @@ get-most-recent-pr-commit() {
 
   curl \
   -sSL \
-  --user deis-admin:"\${GITHUB_ACCESS_TOKEN}" \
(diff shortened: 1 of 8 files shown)


bash/scripts/commit_description_parser.sh
```diff
@@ -49,7 +49,7 @@ get-most-recent-pr-commit() {
 
   curl \
   -sSL \
-  --user deis-admin:"\${GITHUB_ACCESS_TOKEN}" \
+  --user deis-admin:"${GITHUB_ACCESS_TOKEN}" \
```
vdice, line 52: @helgi I believe this was the "derp" change needed (used to escaping `$` in groovy scripts in this repo but we do not want to escape it in the bash environment).  Still, I think the other refactor change in this PR is worthwhile (placing credentials ID in one location whenever it needs to be updated in the future).  If agreed, let's move this through.  Otherwise, I can close this and only PR the change in this file.

PR #142, opened 2016-07-12
fix(commit parser): fix parsing logic and add debug output

vdice:
cc @helgi @sgoings

### DIFF
--- a/bash/scripts/commit_description_parser.sh
+++ b/bash/scripts/commit_description_parser.sh
@@ -6,6 +6,7 @@ main() {
   if [ -n "${JENKINS_HOME}" ]; then
     envPropsFilepath="${WORKSPACE}/env.properties"
   fi
+
   parse-commit-description "${ghprbPullLongDescription}" >> "${envPropsFilepath}"
 }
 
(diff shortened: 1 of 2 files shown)


vdice:
Live example ran in https://ci.deis.io/job/workflow-e2e-pr/416/console

The crucial bit is in the downstream job (https://ci.deis.io/job/workflow-test-pr/3367/console):

```
13:42:57 Setting CONTROLLER_GIT_TAG to git-3efde90
13:42:57 Setting WORKFLOW_E2E_GIT_TAG to git-15ea76d
```

Where `CONTROLLER_GIT_TAG` was inferred from the `Requires ...` in upstream job commit description.

bash/scripts/commit_description_parser.sh
```diff
@@ -16,42 +17,50 @@ parse-commit-description() {
...
-  curl \
+  resp=$(curl \
   -sSL \
   --user deis-admin:"${GITHUB_ACCESS_TOKEN}" \
-  "https://api.github.com/repos/deis/${repoName}/pulls/${prNumber}/commits" \
-  | docker run -i --rm kamermans/jq '.[].sha' | tail -n1
+  "https://api.github.com/repos/deis/${repoName}/pulls/${prNumber}/commits")
+
+  echo "${resp}" >&2
+  echo "${resp}" | docker run -i --rm kamermans/jq '.[].sha'
```
sgoings, line 63: You didn't want to use `last`?

vdice, line 63: @sgoings the return value from that `jq` command was actually a string, albeit multi-lined if more than one `sha` was found... therefore `last` wasn't the appropriate choice.

PR #143, opened 2016-07-12
feat(component_jobs): use PR commit sha for docker tag

vdice:
Rather than merge commit sha.  This enables e2e testing of
multiple 'sibling'/dependent PRs.

cc @helgi @sgoings

### DIFF
--- a/bash/scripts/get_actual_commit.sh
+++ b/bash/scripts/get_actual_commit.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# Gets actual commit for reporting status and potentially tagging Docker image
+main() {
+  envPropsFilepath="/dev/null"
+  if [ -n "${JENKINS_HOME}" ]; then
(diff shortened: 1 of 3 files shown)


vdice:
@sgoings, @helgi this was my idea for fully enabling PR test dependencies.  

Prior to this, on each component build job, the docker artifact produced was always derived from the merge commit, rather than actual PR commit.  Meaning, the build job for PR commit `abc1234` and its corresponding merge commit `def4567` would deploy a new Docker image with tag `git-def4567` on success.

With this change, we make sure the actual PR commit of `abc1234` is used for the Docker tag (so, `git-abc1234`).  Therefore, other PRs dependent on said PR can find the artifacts they need, as the logic finalized in https://github.com/deis/jenkins-jobs/pull/142 simply curls the GitHub API and grabs the actual commit from a given PR number for any required PR.  

Otherwise, any jobs not kicked off by the GitHub Pull Request Builder Plugin (so, manually run or jobs from master merges) are not affected (will defer to versions derived by `make` targets as usual.)

Thoughts?

PR #145, opened 2016-07-13
fix(get_actual_commit.sh): pass down appropriate commit if PR

vdice:
with the actual PR commit (rather than merge commit) so that this
is used by downstream e2e for locating appropriate Docker tag.

cc @helgi

### DIFF
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -100,6 +100,21 @@ repos.each { Map repo ->
         shell new File("${WORKSPACE}/bash/scripts/get_actual_commit.sh").text
 
         shell """
+        #!/usr/bin/env bash
+
+        set -eo pipefail
+
+        # pass along appropriate commit via properties file
+        if [ -z "\${ghprbActualCommit}" ]; then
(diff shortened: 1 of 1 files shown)


vdice:
@helgi live run can be seen here: https://ci.deis.io/job/workflow-test-pr/3433/console (note: `workflow-e2e` image can't be found as corresponding build job needs to be re-run as of https://github.com/deis/jenkins-jobs/pull/144)

jobs/component_jobs.groovy
```diff
@@ -100,6 +100,21 @@ repos.each { Map repo ->
         shell new File("${WORKSPACE}/bash/scripts/get_actual_commit.sh").text
 
         shell """
+        #!/usr/bin/env bash
+
+        set -eo pipefail
+
+        # pass along appropriate commit via properties file
+        if [ "\${ghprbActualCommit}" != "" ]; then
```
jchauncey: could we do a -z test here?

vdice: ah good call; changed

PR #146, opened 2016-07-13
feat(workflow_cli_release.groovy): add workflow-cli-tag-release job

### LINKED ISSUE #135: workflow-cli-release job
Unfortunately, it doesn't appear that a Jenkinsfile-based pipeline is able to trigger a build off of a git tag (see https://issues.jenkins-ci.org/browse/JENKINS-34395).  Therefore, at this time, we require a `workflow-cli-release` job that will act similarly to the existing `<component>_release` [job](https://github.com/deis/jenkins-jobs/blob/master/jobs/component_release.groovy) in that it watches the current `release-${WORKFLOW_RELEASE}` branch for push events and then triggers a build/deployment of the appropriate release artifact -- in this case the `workflow-cli` binary/binaries.

TODO
- [ ] PR in https://github.com/deis/deisrel to add `workflow-cli` to the list used in the creation of release branches (see https://github.com/deis/deisrel/blob/92670ea2b94df2c71010eab4436092be3b233437/actions/common.go#L54 and ensure these are used in said action)

vdice:
Closes #135

### DIFF
--- a/jobs/workflow_cli_release.groovy
+++ b/jobs/workflow_cli_release.groovy
@@ -1,0 +1,83 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+repoName="workflow-cli"
+
+job("${repoName}-tag-release") {
+  description """
+    <li>Watches the ${repoName} repo for a new tag to be pushed. (Assumed to be '\${RELEASE}')</li>
+    <li>It then checks out the \${RELEASE} tag and builds and deploys binaries, optionally kicking off e2e (default: false)</li>
+  """.stripIndent().trim()
+
+  scm {
+    git {
+      remote {
+        github("deis/${repoName}")
+        credentials('597819a0-b0b9-4974-a79b-3a5c2322606d')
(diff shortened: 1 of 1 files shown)


vdice:
So this is a tricky one.  It can't be _wholly_ tested as is, since the last release tag (`v2.1.0`) does not contain two of the necessary `make` targets that are currently in master (`build-revision` and `upload-gcs`).  

_However_, I was able to test this via two different runs.  [One](https://ci.deis.io/job/workflow-cli-tag-release/2/console) verifying that checking out  the current release tag works (logic in `git` block) and [one](https://ci.deis.io/job/workflow-cli-tag-release/5/console) checking out master and verifying that the aforementioned `make` targets work (logic in `shell` block).

We will still want to keep an eye on this job to make sure it runs properly during the next release (`v2.2`) so perhaps I should add a note/reminder in the workflow release checklist?

mboersma:
> perhaps I should add a note/reminder in the workflow release checklist?

Couldn't hurt, although not necessary to merge this IMHO.

PR #147, opened 2016-07-15
fix(common.groovy): check if WORKFLOW_CLI_SHA empty

vdice:
before overriding `CLI_VERSION`

Ref https://github.com/deis/workflow-e2e/pull/260

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -28,7 +28,10 @@ set -eo pipefail
 
 export WORKFLOW_CHART="workflow-${RELEASE}"
 export WORKFLOW_E2E_CHART="workflow-${RELEASE}-e2e"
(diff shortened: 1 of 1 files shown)


common.groovy
```diff
@@ -28,7 +28,10 @@ set -eo pipefail
 
 export WORKFLOW_CHART="workflow-${RELEASE}"
 export WORKFLOW_E2E_CHART="workflow-${RELEASE}-e2e"
-export CLI_VERSION="${WORKFLOW_CLI_SHA:0:7}"
+
+if [ "${WORKFLOW_CLI_SHA}" != "" ]; then
```
jchauncey: could we do a `-n` check here?

jchauncey:
1 small question but LGTM

PR #150, opened 2016-07-19
feat(apptypes): add dockerfile and buildpack jobs

smothiki:
Add nightly jobs to test buildpack and dockerfile apps

### DIFF
--- a/jobs/apptypes_test.groovy
+++ b/jobs/apptypes_test.groovy
@@ -1,0 +1,63 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+import utilities.StatusUpdater
+
+
+name = "apptypes_e2e"
+
+
+  job(name) {
+    description """
+      Runs the buildpack and dockerfile sub-suites of e2e tests on a nightly basis
+    """.stripIndent().trim()
+
+    scm {
+      git {
+        remote {
+          github("deis/e2e-runner")
+        }
+        branch('master')
+      }
+    }
+
+    logRotator {
+      daysToKeep defaults.daysToKeep
+    }
+
+    parameters {
+      stringParam('RELEASE', "dev", "Release string for resolving workflow-[release](-e2e) charts")
+      stringParam('HELM_REMOTE_REPO', defaults.helm["remoteRepo"], "The remote repo to use for fetching charts.")
+      stringParam('E2E_RUNNER_IMAGE', 'quay.io/deisci/e2e-runner:canary', "The e2e-runner image")
+      stringParam('E2E_DIR', '/home/jenkins/workspace/$JOB_NAME/$BUILD_NUMBER', "Directory for storing workspace files")
+      stringParam('E2E_DIR_LOGS', '${E2E_DIR}/logs', "Directory for storing logs. This directory is mounted into the e2e-runner container")
+    }
+
+    triggers {
+      cron('@daily')
+    }
+
+    wrappers {
+      timestamps()
+      colorizeOutput 'xterm'
+      credentialsBinding {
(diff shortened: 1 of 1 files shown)


jobs/apptypes_test.groovy
```diff
@@ -0,0 +1,61 @@
...
+        #!/usr/bin/env bash
+        set -eo pipefail
+
+        echo TEST="bps" >> ${defaults.envFile}
+        echo COMPONENT_REPO="e2e-runner" >> ${defaults.envFile}
+        echo UPSTREAM_SLACK_CHANNEL="testing" >> ${defaults.envFile}
+      """.stripIndent().trim()
+
+      downstreamParameterized {
+        trigger(apptype_e2e) {
```
vdice: this should be the name of the downstream job.  Here we probably want `workflow-test`, right?

Alternately, you could remove the `downstreamParameterized` block and just add another `shell` step calling the default `E2E_RUNNER_JOB` script defined [here](https://github.com/deis/jenkins-jobs/blob/master/common.groovy#L26-L41):

```
shell E2E_RUNNER_JOB
```

jobs/apptypes_test.groovy
```diff
@@ -0,0 +1,56 @@
...
+
+import utilities.StatusUpdater
+
+
+name = "apptypes_e2e"
+
+
+  job(name) {
+    description """
+      Watches the e2e-runner repo for a ${config.type} commit
```
vdice: `config.type` won't be defined... should just change the description here to what this new job will do

jobs/apptypes_test.groovy
```diff
@@ -0,0 +1,51 @@
...
+
+import utilities.StatusUpdater
+
+
+name = "apptypes_e2e"
+
+
+  job(name) {
+    description """
+      runs a e2e-runner job nightly
```
vdice: nit: I'd mention a bit more about how this would just run the `"bps"` test of the e2e suite and what that entails

smothiki: sure will update

smothiki:
https://github.com/deis/workflow-e2e/pull/263
https://github.com/deis/charts/pull/309

jobs/apptypes_test.groovy
```diff
@@ -0,0 +1,63 @@
...
+
+import utilities.StatusUpdater
+
+
+name = "apptypes_e2e"
+
+
+  job(name) {
+    description """
+      runs a e2e-runner job nightly we have to setup TEST=bps env variable for the docker container to execute the test
```
vdice: Can we improve this job description so that the purpose is more readily understood?  Meaning, instead of going into the particulars of `setup TEST=bps env variable for the docker container…` how about describing it more generally, something like `Runs the buildpack and dockerfile sub-suites of e2e tests on a nightly basis`

vdice:
Also @smothiki remember to remove https://ci.deis.io/job/apptypes_test_seed_job once it is no longer needed.

jobs/apptypes_test.groovy
```diff
@@ -0,0 +1,63 @@
...
+        branch('master')
+      }
+    }
+
+    logRotator {
+      daysToKeep defaults.daysToKeep
+    }
+
+    parameters {
+      stringParam('RELEASE', defaults.workflow.release, "Release string for resolving workflow-[release](-e2e) charts")
```
vdice: @smothiki should we change `defaults.workflow.release` to `dev`?  The former will pickup the current/most recent release chart while the latter will pick up, of course, the latest `dev` variant.

vdice:
@smothiki it doesn't look like the `TEST="bps"` env var was used in the most recent run of this job: https://ci.deis.io/job/apptypes_e2e/12/console

smothiki:
refer https://github.com/deis/workflow-e2e/pull/263

PR #160, opened 2016-07-27
feat(clusterator): Add create and delete cluster jobs

jchauncey:


jobs/clusterator_jobs.groovy
```diff
@@ -0,0 +1,67 @@
...
+      string("GCLOUD_CREDENTIALS", "15122437-45b8-48ac-bb84-652394c8a927")
+    }
+  }
+
+  steps {
+    shell """
+      #!/usr/bin/env bash
+
+      set -eo pipefail
+      docker run -e GCLOUD_CREDENTIALS=\${GCLOUD_CREDENTIALS} quay.io/jchauncey/clusterator:git-d3bb8c6
```
vdice: does this command need `create` at the end?

jobs/clusterator_jobs.groovy
```diff
@@ -0,0 +1,67 @@
...
+  triggers {
+    cron('0 7 1 * 1-5')
+  }
+
+  parameters {
+    stringParam('NUMBER_OF_CLUSTERS', '10', 'Number of clusters to create at 1 time')
+    stringParam('NUM_NODES', '5', 'Number of nodes in each cluster')
+    stringParam('MACHINE_TYPE', 'n1-standard-4', 'Node type')
+    stringParam('VERSION', '1.2.5', 'The version of kubernetes to use.')
+    stringParam('sha1', 'master', 'Specific Git SHA to test')
```
vdice: Is this param being used?

jobs/clusterator_jobs.groovy
```diff
@@ -0,0 +1,67 @@
...
+      string("GCLOUD_CREDENTIALS", "15122437-45b8-48ac-bb84-652394c8a927")
+    }
+  }
+
+  steps {
+    shell """
+      #!/usr/bin/env bash
+
+      set -eo pipefail
+      docker run -e GCLOUD_CREDENTIALS=\${GCLOUD_CREDENTIALS} quay.io/jchauncey/clusterator:git-d3bb8c6 delete
```
vdice: Do we want the ability to override the `clusterator` image tag?  And/or would we rather use `canary`/`latest` here (after explicitly running `docker pull` beforehand to bust the cache)?

jobs/clusterator_jobs.groovy
jchauncey: no create is the default

jobs/clusterator_jobs.groovy
jchauncey: no forgot to remove it

jobs/clusterator_jobs.groovy
jchauncey: my thought was to use the immutable image instead of latest

jchauncey: I want to use immutable images if possible.

PR #164, opened 2016-07-28
fix(clusterator): Update cron to run delete job every day at 7pm

jchauncey:


Joshua-Anderson:
What sort of crontab format is this? `H` isn't a valid character in standard crontabs.

jchauncey:
@Joshua-Anderson its valid in Jenkins. It basically says run at any minute during that hour. Helps wiht load.

Joshua-Anderson:
Cool. Good to know!

Issue #155, opened 2016-07-26
Add component image deploy to prod on master merge


vdice:
The current component pipeline on a merge to master consists of two jobs:
1. build, test and deploy component Docker image with `git-<sha>` tag to 'deisci' org of registries
2. run e2e against the workflow-dev chart b/w component tag updated from `1.` above

We would like to add a third:
1. on e2e success, retag image from `1.` into the 'deis'/prod org of registries

This would then provide vetted images that are ready to be released (or at least ready to be vetted via TBD release process and eventually re-tagged with the 'pretty'/semver tag in-place)

Issue #158, opened 2016-07-26
[Meta] <component>-tag-release job


vdice:
As part of the initiative to release a single component at a time (independent of Workflow as a whole) we require a `<component>-tag-release` job for each, which should:
1. watch for git push events to a refspec of `+refs/tags/*:refs/remotes/origin/tags/*` and branch `*/tags/*` (in essence, a tag push event)
2. determine if branch from above is indeed the latest tag, if so, proceed; if no, exit.
3. based on commit at HEAD of branch, find the correlating `git-<commit:0:7>` image tag (ref https://github.com/deis/jenkins-jobs/issues/155) in the 'deis' org of quay and use this for running/vetting in e2e tests
4. assuming e2e passes, retag image from above with 'pretty'/semver release tag (based on actual git tag pushed) in the quay.io and dockerhub 'deis' orgs

TODO for an iterative approach
- [x] create job covering steps 1. and 2. (https://github.com/deis/jenkins-jobs/pull/168)
- [x] promote prod-candidate images on master merge https://github.com/deis/jenkins-jobs/issues/155
- [x] once the e2e strategy is finalized, add step 3. and 4. to job (https://github.com/deis/jenkins-jobs/issues/169)

ref https://github.com/deis/workflow/pull/410

mboersma:
> determine if branch from above is indeed the latest tag, if so, proceed; if no, exit.

Can't this step just build regardless of whether the tag is latest or not? That would cover the case where an older tag is intentionally force-pushed (shouldn't happen, but I'm not sure why we have to ignore it).

mboersma:
Step 3) implies that <component>-release jobs are already publishing prod-candidate images to the deis/ org on success. Is that a separate task or done already?

vdice:
> Step 3) implies that -release jobs are already publishing prod-candidate images to the deis/ org on success. Is that a separate task or done already?

@mboersma good point; description updated linking to task

vdice:
> Can't this step just build regardless of whether the tag is latest or not?

@mboersma it could.  Note however that on a tag push event, our test job in https://ci.deis.io/job/controller-tag-release ran 3 times, one checking out `origin/tags/v2.0.0`, another `origin/tags/v2.1.0` and the last `origin/tags/v2.1.1` (the last being the actual most recent tag)  because of our branch wildcard specifier `*/tags/*`.  

[...]

PR #165, opened 2016-07-28
feat(component_promote): add component-promote job

vdice:
Successful test run in https://ci.deis.io/job/workflow-test-clone-for-issue-155/38/console (which kicked off downstream `component-promote` job https://ci.deis.io/job/component-promote/6/console)

Closes https://github.com/deis/jenkins-jobs/issues/155

### DIFF
--- a/jobs/component_promote.groovy
+++ b/jobs/component_promote.groovy
@@ -1,0 +1,76 @@
+evaluate(new File("${WORKSPACE}/common.groovy"))
+
+job('component-promote') {
+  description """
+    Promotes a component image to the production 'deis' registry org on e2e success after a merge to master
+  """.stripIndent().trim()
+
+
+  concurrentBuild()
+  throttleConcurrentBuilds {
+    maxPerNode(defaults.maxBuildsPerNode)
+    maxTotal(defaults.maxTotalConcurrentBuilds)
+  }
+
+  logRotator {
+    daysToKeep defaults.daysToKeep
+  }
+
+  publishers {
+    slackNotifications {
+      notifyFailure()
+      notifySuccess()
+    }
+  }
+
+  parameters {
+    stringParam('DOCKER_USERNAME', 'deisbot', 'Docker Hub account name')
+    stringParam('DOCKER_EMAIL', 'dummy-address@deis.com', 'Docker Hub email address')
+    stringParam('QUAY_USERNAME', 'deis+jenkins', 'Quay account name')
+    stringParam('QUAY_EMAIL', 'deis+jenkins@deis.com', 'Quay email address')
+    stringParam('COMPONENT_NAME', '', 'Component name')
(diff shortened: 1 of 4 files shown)


jobs/component_promote.groovy
```diff
@@ -0,0 +1,60 @@
...
+
+      set -eo pipefail
+
+      image_name_and_tag="${COMPONENT_NAME}:git-${COMPONENT_SHA:0:7}"
+      original_image=quay.io/deisci/"${image_name_and_tag}"
+      promoted_image=quay.io/deis/"${image_name_and_tag}"
+
+      echo "Promoting ${original_image} to ${promoted_image}"
+
+      docker login -e="\$QUAY_EMAIL" -u="\$QUAY_USERNAME" -p="\$QUAY_PASSWORD" quay.io
```
jchauncey: My only question is should we do this for both quay and dockerhub?

jchauncey:
This is great. My only question is should we be publishing these images to both quay and dockerhub?

jobs/component_promote.groovy
vdice: @jchauncey I had discussed this with @mboersma offline and we came to the conclusion that just pushing to the quay registry was sufficient since `quay.io/deis/...` images are the ones listed in the release chart.  Meaning, once the component release pipeline runs, it will look up the image promoted here (to quay), sub that in for the most recent semver release chart, run tests, and then, finally, retag with the official/'pretty' semver tag to both quay and dockerhub.

That being said, it would be easy to add dockerhub here, too.  I'm definitely open to that...

jchauncey:
my concern is that in the case where quay is down when someone is doing an install we have no fall back for them. It would be great if we could say oh just use dockerhub or something like that. Not that quay goes down all that often but last year we had a 18 hour outage with quay that meant I could do no deploys. That isn't great =\

vdice:
@jchauncey that makes sense -- but if someone doing an install of a release chart, the tags referenced will be the official semver tags, which, as mentioned, will live in both quay and dockerhub (see `4.` in https://github.com/deis/jenkins-jobs/issues/158).  Otherwise, the user would be installing the 'dev' chart in which case only quay currently holds the 'canary' tags... and I don't think we have plans to to change that.

But again, I can add dockerhub here... I'll defer to your judgement.

jchauncey:
yeah but we will have to push the image to dockerhub whereas with quay we can just retag. It is probably worth doing that operation up front to save us time at release

jchauncey:
@mboersma what do you think?

vdice:
@jchauncey that's a good point on efficiency of retagging.  I'll start adding dockerhub as well, unless @mboersma makes the call otherwise.

vdice:
updated PR to push to dockerhub as well; ready for review

PR #173, opened 2016-07-29
chore(common): bump limit for workflow-test-pr

Joshua-Anderson:
We now have 16 clusters available on weekdays, so the number needs a bump.

Bonus points for anyone who knows how to create a followup PR the only raises the cap when clusterator is active. 

Ping @jchauncey

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -57,7 +57,7 @@ defaults = [
   maxBuildsPerNode: 1,
   maxTotalConcurrentBuilds: 3,
   maxWorkflowTestConcurrentBuilds: 3,
-  maxWorkflowTestPRConcurrentBuilds: 4,
+  maxWorkflowTestPRConcurrentBuilds: 13,
   maxWorkflowReleaseConcurrentBuilds: 1,
   workflow: [
     chartName: 'workflow-dev',


common.groovy
```diff
@@ -56,7 +56,7 @@ defaults = [
   ],
   maxBuildsPerNode: 1,
   maxTotalConcurrentBuilds: 3,
-  maxWorkflowTestConcurrentBuilds: 3,
+  maxWorkflowTestConcurrentBuilds: 13,
```
vdice: if this PR is meant for `workflow-test-pr`, I believe we would bump the value immediately below this one.  (This one I believe is for `workflow-test`)

Joshua-Anderson:
@vdice Fixed!

jchauncey:
That would be easy to do but requires us to run the seed job all the time

jchauncey:
@Joshua-Anderson this should be fine. I was actually thinking about just removing the limit completely

Joshua-Anderson:
I wouldn't entirely remove the limit because it is possible to queue up a lot of builds at once. Merging a SDK or CLI PR kicks off a rebuild of all the pull requests in the repository. If those are in high enough numbers, it can overwhelm the cluster capacity.

jchauncey:
We could probably up the number of clusters we create to handle that case.

helgi:
Why does the SDK / CLI need to rebuild all PRs?

Joshua-Anderson:
A quirk (or is it a feature 😛) of the jenkinsfile build system is that instead of building the commit of the PR, it builds the commit of the pull request merged into master that github would generate if you merged the PR using the web UI. When somebody merges a PR, this merge commit changes and jenkins kicks off a new PR build to make sure the PR is passing with the new commit.

For reference Travis CI also uses the merge commit to build PRs, but it doesn't rebuild if a PR is merged.

PR #178, opened 2016-08-01
ref(component_tag_release.groovy): revisit release pipeline

vdice:
Instead of hingeing on a git webhook trigger (which has proven faulty/overzealous
in running on webhooks other than the latest tag push event), refactor
pipeline to kick off manually with provided TAG value.  The new pipeline will then:
1. locate release candidate off of latest commit in master
2. if successful (meaning that commit has been vetted by e2e), run 'git tag TAG' commands
3. kick off e2e against release candidate
4. if successful, retag release candidate with TAG in docker registries


bacongobbler:
I'm okay with this process for the short term. 👍

PR #186, opened 2016-08-03
feat(commit_description_parser.sh): get commit description from git

vdice:
Rather than depending on ghprb plugin, so that the commit description can be used by retrying the job or kicking it off manually (neither of which trigger the ghprb plugin)


vdice:
Tested via https://ci.deis.io/job/workflow-e2e-pr-clone/3/console (see also the downstream e2e job https://ci.deis.io/job/workflow-test-pr/4674/console)

PR #189, opened 2016-08-03
fix(workflow_cli_release): update gcs upload strategy.

Joshua-Anderson:
cc @vdice

### DIFF
--- a/jobs/workflow_cli_release.groovy
+++ b/jobs/workflow_cli_release.groovy
@@ -48,17 +48,23 @@ job("${repoName}-tag-release") {
   }
 
   steps {
-    shell '''
+    gcloud = "docker run --rm -v  /tmp/workflow-cli-release:/.config -v \${WORKSPACE}/_dist:/upload google/cloud-sdk:latest"
+    gcs_bucket = "gs://workflow-cli"
+    shell """
       #!/usr/bin/env bash
 
       set -eo pipefail
 
-      mkdir -p ${WORKSPACE}/tmp
+      mkdir -p /tmp/workflow-cli-release
 
-      cat "${GCSKEY}" > ${WORKSPACE}/tmp/key.json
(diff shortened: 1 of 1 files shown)


jobs/workflow_cli_release.groovy
```diff
@@ -57,7 +59,11 @@ job("${repoName}-tag-release") {
 
       cat "${GCSKEY}" > ${WORKSPACE}/tmp/key.json
 
-      GIT_TAG="${RELEASE}" make bootstrap build-revision upload-gcs
+      GIT_TAG="${RELEASE}" make bootstrap build-revision fileperms
+
+      ${gcloud} gcloud auth activate-service-account -q --key-file /.config/key.json'
```
vdice: now that we are referencing groovy vars like `gcloud` (and want interpolation performed), we will need to change the `'''` surrounding this shell step with `"""`, as well as escaping the `$` before any non-groovy (so, bash environment) vars so they don't get interpolated.  (for instance, `"\${RELEASE}"`, etc.)

jobs/workflow_cli_release.groovy
```diff
@@ -48,17 +48,23 @@ job("${repoName}-tag-release") {
...
+      mkdir -p \${WORKSPACE}/tmp
 
-      cat "${GCSKEY}" > ${WORKSPACE}/tmp/key.json
+      cat "\${GCSKEY}" > \${WORKSPACE}/tmp/key.json
 
-      GIT_TAG="${RELEASE}" make bootstrap build-revision upload-gcs
-    '''.stripIndent().trim()
+      GIT_TAG="\${RELEASE}" make bootstrap build-revision fileperms
+
+      ${gcloud} gcloud auth activate-service-account -q --key-file /.config/key.json'
```
vdice: there is an errant `'` char at the end of this line it appears.

PR #199, opened 2016-08-09
fix(component_jobs.groovy): only parse commit description if pr job

vdice:
cc @jackfrancis

### DIFF
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -132,7 +132,8 @@ repos.each { Map repo ->
           DEIS_REGISTRY=quay.io/ make docker-build ${dockerPush}
         """.stripIndent().trim()
 
-        shell new File("${WORKSPACE}/bash/scripts/commit_description_parser.sh").text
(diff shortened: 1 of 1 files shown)


jobs/component_jobs.groovy
```diff
@@ -132,7 +132,8 @@ repos.each { Map repo ->
           DEIS_REGISTRY=quay.io/ make docker-build ${dockerPush}
         """.stripIndent().trim()
 
-        shell new File("${WORKSPACE}/bash/scripts/commit_description_parser.sh").text
+        // only attempt to parse commit description if -pr job
+        isMaster ?: shell new File("${WORKSPACE}/bash/scripts/commit_description_parser.sh").text
```
jackfrancis: I assume the absence of anything between the `?` and the `:` means a `noop` for the truthy path of this ternary expression?

vdice: @jackfrancis correct.  http://docs.groovy-lang.org/latest/html/documentation/index.html#_elvis_operator

PR #208, opened 2016-08-19
ref(*): refactor bash logic

### LINKED ISSUE #193: env.properties file not found
Sometimes e2e will pass successfully but Jenkins can't find the `env.properties` file to read from/write to.  See https://ci.deis.io/job/workflow-test/1458/console

It _seems_ to happen only for `<job>@<#>` job workspaces (as opposed to non-concurrent `<job>` workspace).  Therefore, it most definitely is a race when a job is running concurrently on the same node as another.  We should be writing/reading from a file in `/tmp/` with each job's unique `BUILD_NUMBER` on each node à la the default [here](https://github.com/deis/jenkins-jobs/blob/master/common.groovy#L49).

vdice:
Make scripts more dynamic so that args can be passed to them
and values can be returned, instead of static scripts that
communicate via env files/variables.

To that end, when an env file is necessary, this makes sure either to use the `defaults.envFile` as defined in [common.groovy](https://github.com/deis/jenkins-jobs/blob/master/common.groovy#L28) _or_ the component-specific env file as defined per component in [repo.groovy](https://github.com/deis/jenkins-jobs/blob/master/repo.groovy#L92).

Additionally, I removed unused imports of the `StatusUpdater` utility, fixed spacing where needed and otherwise cleaned up most files.

Closes https://github.com/deis/jenkins-jobs/issues/193


vdice:
This is now ready for review; still marking it for manual testing (meaning running some pipelines w/ jobs touched here) which I hope to get to in the near future.

vdice:
Tested pipelines manually off of this branch again, didn't find any issues.  Looking for reviews and getting this into use.

Issue #206, opened 2016-08-17
workflow-cli-tag-release job not catching the tag push webhook


vdice:
Whereas the other `<component>-release` jobs are (see `component_release.groovy`).  Therefore, will probably want to match that `scm` pattern as a first fix.

PR #213, opened 2016-08-24
fix(workflow_cli_release.groovy): update job to properly catch git tag push

vdice:
Update this job to use proven working git tag-triggered logic in jobs created by `component_release.groovy`.

Also remove the `RUN_E2E` step/logic as it is not used.

Fixes https://github.com/deis/jenkins-jobs/issues/206

TODO:
- [x] delete old/re-named `workflow-cli-tag-release` [job](https://ci.deis.io/job/workflow-cli-tag-release/)

### DIFF
--- a/jobs/workflow_cli_release.groovy
+++ b/jobs/workflow_cli_release.groovy
@@ -2,10 +2,10 @@ evaluate(new File("${WORKSPACE}/common.groovy"))
 
 repoName="workflow-cli"
 
-job("${repoName}-tag-release") {
+job("${repoName}-release") {
   description """
-    <li>Watches the ${repoName} repo for a new tag to be pushed. (Assumed to be '\${RELEASE}')</li>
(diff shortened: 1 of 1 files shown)


jobs/workflow_cli_release.groovy
```diff
@@ -31,15 +31,15 @@ job("${repoName}-tag-release") {
   }
 
   parameters {
-    stringParam('RELEASE', defaults.workflow.release, 'Release to use for tag checkout')
-    booleanParam('RUN_E2E', false, 'check to run downstream release e2e job')
+    stringParam('TAG', '', 'Specific tag to release')
```
Joshua-Anderson, line 34: Does this do anything? I don't see it used anywhere?

vdice: @Joshua-Anderson good question as it's definitely not obvious from this change set.  The `TAG` env var is inspected in the previously-existing `get_latest_tag.sh` [script](https://github.com/deis/jenkins-jobs/blob/master/bash/scripts/get_latest_tag.sh), the use of which was added in this PR.

Issue #151, opened 2016-07-21
Enable 'Requires' for a deis/charts PR


vdice:
Currently, although the `Requires` [logic](https://github.com/deis/jenkins-jobs/blob/master/bash/scripts/commit_description_parser.sh) identifies a PR number in any repo in the `deis` org (and will find the commit sha for that PR) -- it only actually utilizes said commit if it is one of the Workflow components with env vars listed [here](https://github.com/deis/jenkins-jobs/blob/master/common.groovy#L1-L21) or [here](https://github.com/deis/jenkins-jobs/blob/master/jobs/workflow_test.groovy#L92)

It would be useful to be able to actually use a commit found from a PR in the `deis/charts` repo, although this may involve quite a bit of logic.  For instance, what do we do with said commit?  It isn't a component that can be reduced to a Docker image artifact -- rather, the commit represents changes to one or more charts.  

If the changed charts are the usual `workflow-dev(-e2e)` charts, then it may not be too hard to just checkout the [forked repo](https://github.com/deis/e2e-runner/blob/master/scripts/config.sh#L8) at the given [sha or branch](https://github.com/deis/e2e-runner/blob/master/scripts/config.sh#L10) and use those versions of the charts to run the standard downstream e2e job.  Indeed, this may be the only case needed.

Otherwise, if the PR presents new or different charts for Workflow and E2E, we would need to be able to determine the names from the PR doing the `Requires` and sending that info down to the appropriate spot(s), eventually `deis/e2e-runner`, which does provide the ability to override the names of both charts via [env vars](https://github.com/deis/e2e-runner/blob/master/scripts/config.sh#L12-L13).

PR #215, opened 2016-08-24
feat(*): enable requiring a deis/charts commit

vdice:
We already are set up to parse a required PR in any deis repo
provided it is mentioned in the git commit body, e.g. `Requires deis/<repo>#<pr_number>`

(Previously, we only actually performed subsequent logic if it was a Workflow component repo.)

This sets us up so that we can perform logic if the required PR is in the (presumably forked) deis/charts repo.  

This logic consists of determining the appropriate repo clone url alongside the charts commit sha and passing this info down do the downstream e2e job so that it can set up its local helmc instance appropriately, before fetching charts.

Closes https://github.com/deis/jenkins-jobs/issues/151

TODO: manually test the added logic in this PR by:
- [x] running the branch through Jenkins to be sure it parses correctly
- [x] create a `workflow-test`-clone with logic hingeing off `CHARTS_SHA`
- [x] issuing a test pr in a Workflow component repo requiring a PR in charts, with this logic running on Jenkins
- [ ] delete https://ci.deis.io/job/require-charts-pr-seed-job/ once this is merged

### DIFF
--- a/bash/scripts/find_required_commits.sh
+++ b/bash/scripts/find_required_commits.sh
@@ -4,8 +4,8 @@ set -eo pipefail
 # find-required-commits parses a commit description for any required sibling
 # commits to pass along to downstream job(s)...
 find-required-commits() {
-  local git_commit="${1}"
-  local commit_description="${COMMIT_DESCRIPTION:-$(git log --format=%B -n 1 "${git_commit}")}"
+  git_commit="${1}"
+  commit_description="${COMMIT_DESCRIPTION:-$(git log --format=%B -n 1 "${git_commit}")}"
 
   # Looks specifically for matches of '[rR]equires <repo>#<pr_number>',
   # e.g., "requires builder#123, Requires router#567"
--- a/bash/tests/run_e2e_test.bats
+++ b/bash/tests/run_e2e_test.bats
@@ -1,0 +1,50 @@
+#!/usr/bin/env bats
+
+setup() {
+  . "${BATS_TEST_DIRNAME}/../scripts/run_e2e.sh"
+  load stub
+  stub docker
+  stub mkdir
+
+  E2E_DIR="${BATS_TEST_DIRNAME}/tmp"
+  E2E_DIR_LOGS="${E2E_DIR}/logs"
+  RELEASE="dev"
+}
+
+teardown() {
+  rm_stubs
+}
+
+@test "run-e2e : bogus default env file" {
+
+  run run-e2e "bogus/default/env/file"
+
+  [ "${status}" -eq 0 ]
+  [ "${output}" == "" ]
+  [[ -e "${E2E_DIR_LOGS}" ]]
+  [[ "$(cat "${E2E_DIR}/env.file")" == *"CLI_VERSION=latest"* ]]
+}
+
+@test "run-e2e : real default env file" {
+  default_env_file="${BATS_TEST_DIRNAME}/tmp/default_env.file"
+
+  echo "FOO=bar" > "${default_env_file}"
+
+  run run-e2e "${default_env_file}"
+
+  [ "${status}" -eq 0 ]
+  [ "${output}" == "" ]
+  [ -e "${E2E_DIR_LOGS}" ]
+  [[ "$(cat "${E2E_DIR}/env.file")" == *"CLI_VERSION=latest"*"FOO=bar"* ]]
+}
+
+@test "run-e2e : WORKFLOW_CLI_SHA set" {
+  WORKFLOW_CLI_SHA="abc1234def5678"
+
+  run run-e2e "bogus/default/env/file"
+
+  [ "${status}" -eq 0 ]
+  [ "${output}" == "" ]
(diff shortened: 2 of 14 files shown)


vdice:
The logic stemming from the presence of a non-empty `CHARTS_SHA` env var has been tested to success in two cases:
- CHARTS_SHA has already been merged: https://ci.deis.io/job/workflow-test-clone-for-pr-215/15/console
- CHARTS_SHA still in PR form: https://ci.deis.io/job/workflow-test-clone-for-pr-215/16/console

(note in the latter job, e2e failed on account of the PR itself)

All that is left is a quick test to make sure a (dummy) PR with 'Requires charts#<pr number>` gets picked up -- and since that logic already exists, there _should_ be no issues.

vdice:
e2e run of this logic can be seen in https://ci.deis.io/job/workflow-e2e-pr/576/console which kicked off https://ci.deis.io/job/workflow-test-pr/5842/console (note the latter job failed due to actual charts pr logic)

bash/scripts/find_required_commits.sh
```diff
@@ -4,8 +4,8 @@ set -eo pipefail
 # find-required-commits parses a commit description for any required sibling
 # commits to pass along to downstream job(s)...
 find-required-commits() {
-  local git_commit="${1}"
-  local commit_description="${COMMIT_DESCRIPTION:-$(git log --format=%B -n 1 "${git_commit}")}"
+  git_commit="${1}"
```
vdice, line 7: Was following a bad practice of declaring and assigning a `local` var in the same line -- and in fact this broke some logic.  No need for them to be `local` really, so removed.

sgoings: Bad practice per what advice?

vdice: @sgoings http://tldp.org/LDP/abs/html/localvar.html#EXITVALANOMALY01 clued me in to weird behavior that can result -- though that example doesn't seem to be related to our use here, I noticed in the prev version `local var=value` that the finding of required commits wouldn't run correctly (at some point the values weren't available?  I never really found a concrete answer).  When `local` was removed, the logic worked again as expected.

sgoings: Yeah I'm familiar with that behavior. Using local clobbers the return code of the execution of a subshell or the setting of the variable. So if you're relying on grabbing the return code after setting the variable with the command's output you can't prefix the line with local. However, if the subshell errors out, the script will still fail due to the `set -eo pipefail`.

If this solves an actual problem... cool. But I'm still intrigued by what actually might be happening.

bash/tests/run_e2e_test.bats
```diff
@@ -0,0 +1,50 @@
...
+  [[ "$(cat "${E2E_DIR}/env.file")" = *"CLI_VERSION=latest"*"FOO=bar"* ]]
+}
+
+@test "run-e2e : WORKFLOW_CLI_SHA set" {
+  WORKFLOW_CLI_SHA="abc1234def5678"
+
+  run run-e2e "bogus/default/env/file"
+
+  [ "${status}" -eq 0 ]
+  [ "${output}" = "" ]
```
sgoings: I know a single `=` gives you the behavior you want... but considering other languages have different behavior for `=` vs. `==` I'd rather see the `==` here.

jobs/workflow_test.groovy
```diff
@@ -129,20 +130,28 @@ import utilities.StatusUpdater
       if (isPR) { // update commit with pending status while tests run
         shell StatusUpdater.updateStatus(
           commitStatus: 'pending', repoName: '${COMPONENT_REPO}', commitSHA: '${ACTUAL_COMMIT}', description: 'Running e2e tests...')
+
+        setupHelmcEnv = new File("${WORKSPACE}/bash/scripts/setup_helmc_environment.sh").text
+        setupHelmcEnv += """
+          mkdir -p ${defaults.tmpPath}
+          setup-helmc-env >> ${defaults.envFile}
```
sgoings, line 137: Why is this behavior "append" now?

bash/tests/run_e2e_test.bats
vdice: @sgoings prudent call.  Objections to tackling this issue (same `=` in all bats tests currently) in a follow-up ticket/PR?  I can create it now.

jobs/workflow_test.groovy
vdice: It doesn't strictly need to be since we see here this is the first time we are writing to the file -- I've just been in the habit of appending by default since the file is located in a JOB_NAME/BUILD_NUMBER-specific location, just to be 'safe' as there can be many times where we write to it.

That being said, I can certainly change to `>` for correctness.

bash/tests/run_e2e_test.bats
vdice: https://github.com/deis/jenkins-jobs/pull/218

vdice:
@sgoings could you revisit when you have a chance?

PR #219, opened 2016-08-29
fix(get_actual_commit.sh): curl GH API for actual pr commit, if pr

vdice:
cc @helgi 

Previous logic determined the commit parents of a given merge commit (if pr) via `git` and assumed the first parent was the pr commit, based on our experience with same logic in workflow-cli/controller-sdk-go Jenkinsfiles.  Turns out, based on every PR I have since looked up GH info from, in these Jenkins jobs using the Github Pull Request Builder, the _second_ parent is the actual PR commit.  

As an alternative to this obviously faulty approach, I've changed the logic up to curl the canonical source (the GitHub API itself) and grabbing the commit message on a pr `Merge <pr_sha> into <master sha>` and deriving the actual PR commit from there.  Not sure if this is the _best_ way, but it should be an improvement.

### DIFF
--- a/bash/scripts/get_actual_commit.sh
+++ b/bash/scripts/get_actual_commit.sh
@@ -3,18 +3,32 @@
 set -eo pipefail
 
 get-actual-commit() {
+  repo_name="${1}"
   git_branch="${GIT_BRANCH:-$(git describe --all)}"
   git_commit="${GIT_COMMIT:-$(git rev-parse HEAD)}"
 
   if [[ "${git_branch}" != *"master"* ]]; then
-    # Determine actual PR commit for reporting status and Docker image tagging
-    git_commit_parents="${GIT_COMMIT_PARENTS:-$(echo "${git_commit}" | git log --pretty=%P -n 1 --date-order)}"
-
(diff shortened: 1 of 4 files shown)


bash/scripts/get_actual_commit.sh
```diff
@@ -3,18 +3,32 @@
...
+
+  # if pr, commit will be the merge commit with pr commit and master commit as its parents
+  # curl GH api to derive actual pr commit for reporting status and Docker image tagging
+  resp=$(curl \
+  -sSL \
+  --user deis-admin:"${GITHUB_ACCESS_TOKEN}" \
+  "https://api.github.com/repos/deis/${repo_name}/commits/${commit}")
+
+  echo "${resp}" \
+    | docker run -i --rm kamermans/jq '.commit.message' \
```
helgi, line 31: A docker container to run jq?

vdice: yea @helgi , have been doing that since jenkins slaves don't have jq yet, but do have docker.  

I can issue a follow-up pr to get it on the slaves and then refactor the containerized jq out of jenkins-jobs if desired.

PR #222, opened 2016-08-30
ref(bash/scripts/*): use jq directly

vdice:
Instead of via docker image, now that binary is on jenkins slaves

Ref https://github.com/deis/ci-manager/pull/3

cc @helgi


vdice:
updated PR w/ small test-related change -- found easier way to stub a binary dependency in-line rather than going through hassle of generating one via the (now removed) `default.bash` stub.

PR #224, opened 2016-08-31
feat(jobs/clusterator_jobs.groovy): randomize cluster version

vdice:
If VERSION is not set/empty, randomize each created cluster version based on default list provided.

ci/e2e will then be testing on a variety of k8s/GKE versions as the default day-to-day.

### DIFF
--- a/bash/scripts/clusterator_create.sh
+++ b/bash/scripts/clusterator_create.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+clusterator-create() {
+  k8s_versions="${1}"
+
+  if [ -z "${VERSION}" ]; then
+    echo "Assigning random version for each cluster based on provided list:"
+    echo "${k8s_versions}"
+  fi
+
+  clusters_created=0
+  while [ $clusters_created -lt "$NUMBER_OF_CLUSTERS" ]; do
+    cluster_version="${VERSION:-$(echo "${k8s_versions}" | shuf -n 1)}"
+    echo "Creating cluster with version: ${cluster_version}..."
+
+    docker run \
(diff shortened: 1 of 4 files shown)


Joshua-Anderson:
This makes me a little nervous. E2E is unstable enough as it is, I'm afraid we wouldn't be able to connect the dots to a k8s specific failure vs a sporadic failure.

vdice:
@Joshua-Anderson that's definitely a good point.  Would it help if we logged the cluster version in the e2e run logs?  If so, would you consider that a prerequisite before making this change?

Another possibility might be to bring up all clusters w/ latest GKE version as default and then have a different job (nightly?) that ran the whole gamut of:
1. bring up `<other supported GKE version>` cluster (via clusterator)
2. run e2e
3. tear down `<other supported GKE version>` cluster (via clusterator)

The beauty of the approach above is the job itself defines the version so this version can be printed loud and clear on the build label for easy groking...

Joshua-Anderson:
I think nightly tests are a good idea for sure.

Making the k8s version obvious is also going to be a good start.

Basically, we want to make sure people don't just restart e2e until it passes.
I for one probably wouldn't consider the k8s version being the cause if I got a random e2e failure, I'd just restart it and wouldn't notice that it passed when it got a particular version, I'd pin it down to luck. Keeping track of failure rates of the different versions could be interesting and could maybe yield some insight though.

jchauncey:
This is probably ok as long as we specifically tie the workflow-test and workflow-test-pr jobs to a particular version.

vdice:
In considering the helpful feedback above, we've decided to nix this approach in favor of a more deterministic/job-defined approach for testing on clusters w/ GKE/k8s versions other than the latest release; closing.

PR #225, opened 2016-08-31
ref(*): add charts and workflow-cli repos to list

vdice:
But with `buildJobs` bool set to `false`.  Meaning, don't build jobs for them
like the other Workflow components, but do create env vars based on commit
shas for use in other jobs.

(note the only changes in `component_jobs.groovy` and `component_release.groovy` is the check for `repo.buildJobs != false` and resulting indentation of body, contrary to what GH seems to be presenting here.)

### DIFF
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -1,177 +1,179 @@
 evaluate(new File("${WORKSPACE}/common.groovy"))
 
 repos.each { Map repo ->
-  [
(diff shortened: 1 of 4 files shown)


helgi:
You can add `?w=1` to ignore those space changes: https://github.com/deis/jenkins-jobs/pull/225/files?w=1

vdice:
@helgi nice tip; thanks!

PR #226, opened 2016-08-31
feat(*): bring in job-dsl-gradle structure

vdice:
For processing job dsl against versions of plugins spec'd in build.gradle

Added the new make targets of `test-dsl` and `docker-test-dsl` to corresponding
`test` and `docker-test` targets so that ci will test dsl changes on PRs.

This will also be handy for testing upgraded plugin versions against current DSL
before making the leap in Jenkins.

See https://github.com/sheehan/job-dsl-gradle-example

TODOs:
- [x] smoke/regression test the dsl changes by running seed job on this branch and running pipelines
- [x] **post-lgtm:** ~~issue follow-up PR~~ create issue to test upgrading to latest job-dsl-plugin version (1.50 as of writing): https://github.com/deis/jenkins-jobs/issues/231
- [x] **post-lgtm:** ~~issue follow-up PR~~ create issue to extract status updating script into `bash/scripts` and cover under minimal bats test: https://github.com/deis/jenkins-jobs/issues/232

cc @sgoings for review and acknowledgement on kickstarting and helping with this effort!


bacongobbler:
Would you also mind adding documentation? Specifically around the new Makefile targets and how a contributor would use them for testing their changes. :)

vdice:
@bacongobbler Thank you for the reminder!  The updated README is now up for review.

bacongobbler:
Very well written.

build.gradle
```diff
@@ -0,0 +1,73 @@
...
+test {
+    dependsOn tasks.resolveTestPlugins
+    inputs.files sourceSets.jobs.groovy.srcDirs
+
+    // set build directory for Jenkins test harness, JENKINS-26331
+    systemProperty 'buildDirectory', project.buildDir.absolutePath
+}
+
+task wrapper(type: Wrapper) {
+    gradleVersion = '2.14'
```
sgoings: Why not 3.0?

jobs/workflow_test.groovy
```diff
@@ -64,8 +63,20 @@ import utilities.StatusUpdater
...
+                       curl \
+                       --silent \
+                       --user deis-admin:"\${GITHUB_ACCESS_TOKEN}" \
+                       --data '{ \
+                       "state":"${commitStatus}", \
+                       "target_url":"'"\${BUILD_URL}"'", \
+                       "description":"${name} job ${buildStatus}", \
+                       "context":"ci/jenkins/pr"}' \
+                       "https://api.github.com/repos/deis/\${COMPONENT_REPO}/statuses/\${ACTUAL_COMMIT}"
+                     """.stripIndent().trim()
```
sgoings, line 80: Looks like maybe this is part of some conflicts/new stuff that's dropped in master since PR was created? Rebase should help?

sgoings:
YAAAS. This is awesome @vdice.

sgoings:
Manually tested via:

```
make test-dsl
make open-test-results
```

All looks bueno.

⚠️  I'm not sure we'll actually need the `docker-` prefixed make targets (and some might say having make + gradle in the same repo is a bit wonky...).

Your TODO list is 💯 . Glad to see we're getting some better testing around these jenkins jerbs!

build.gradle
vdice: @sgoings good catch -- indeed, why not?!  Updated.

jobs/workflow_test.groovy
vdice: @sgoings thanks - rebased; lookin' good.

vdice:
@sgoings 

> I'm not sure we'll actually need the docker- prefixed make targets (and some might say having make + gradle in the same repo is a bit wonky...).

Yea, I'll definitely defer to you on this (albeit in a subsequent PR ideally) as you possess the greatest Gradle knowledge and experience at this point.  

My thinking for this initial introduction was, as this is our first/only Groovy/Java/Gradle project, that most devs (and Jenkins slaves) would not have the necessary dependencies installed.  (Indeed, I had gone this far at Deis w/o installing Java!)

In addition, as containerized builds/tests are the norm/standard in other repos, I thought it would be fitting to do the same here, so that hopefully the aforementioned devs/contributors could be productive and vet their job contributions in a short amount of time...

sgoings:
> as containerized builds/tests are the norm/standard in other repos

I get that reasoning... I'm not holding this PR up for any reason over this point. 😄

vdice:
Alright, I updated this PR with the following changes:
- Added the rest of the plugins used currently by jobs into the `build.gradle` file
- Added a `publishHTML` step to the `jenkins_jobs_pr.groovy` job so that the test results are published after a ci run
- Re-ran `./gradlew wrapper` to generate an updated gradle wrapper based on the `3.0` version update
- Defaulted `make docker-test-dsl` to running the docker image fresh (meaning, downloading all dependencies anew) which I believe should be the default for ci and added `make docker-test-dsl-quick` that can start a stopped `gradle-test` container for speedier test cycles when job DSL is amended
- Updated README where necessary (mostly the last item mentioned, also describing how this test harness enables vetting plugin additions/upgrades)

I'll ping both reviewers on Tuesday after the holiday to revisit.

vdice:
In other news, the initial approach of attempting to inspect the env var `WORKSPACE` (via `System.getenv('WORKSPACE')`) did not work on Jenkins, as it resolved to null there as well.  After many alternate attempts, the best I could come up with was:

```
def workspace = new File(".").getAbsolutePath()
if (!new File("${workspace}/common.groovy").exists()) { workspace = "${WORKSPACE}"}
evaluate(new File("${workspace}/common.groovy"))
```

Definitely open to nicer/more efficient solution but this works for now...

jobs/workflow_test_release.groovy
```diff
@@ -1,4 +1,6 @@
-evaluate(new File("${WORKSPACE}/common.groovy"))
+def workspace = new File(".").getAbsolutePath()
+if (!new File("${workspace}/common.groovy").exists()) { workspace = "${WORKSPACE}"}
```
bacongobbler: do we want to go a little further and check if the file is readable, or is `.exists()` sufficient enough for our use case?

vdice: @bacongobbler good idea, let's.  Updated to use `canRead()`.

bacongobbler:
LGTM still stands

Makefile
```diff
@@ -1,13 +1,42 @@
...
+
+docker-test-style:
 	${TEST_ENV_PREFIX} ${SHELLCHECK_CMD}
+
+docker-test-scripts:
 	${TEST_ENV_PREFIX} ${BATS_CMD}
 
-.PHONY: test docker-test
+docker-test-dsl:
+	docker rm gradle-test || true
```
sgoings: In makefiles, you want to prefix a line with `-` if you don't care about the error: https://www.gnu.org/software/make/manual/html_node/Errors.html

sgoings:
Ship it once my super ticky tack change is fixed!

PR #230, opened 2016-09-06
ref(workflow_cli_release.groovy): trigger multiple downstream jobs

vdice:
So that on a workflow-cli release event, both the immutable release artifact and updated `deis-stable` artifact
can be built/uploaded concurrently.

TODO:
- [x] test the revised job to be sure it uploads both immutable and mutable release artifacts -- we'll be doing this for the next `v2.5.0` workflow-cli release.

Ref https://github.com/deis/workflow-cli/pull/218

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -1,9 +1,6 @@
 evaluate(new File("${WORKSPACE}/repo.groovy"))
 
-def workflowRelease = [
-  chart: 'v2.5.0',
-  cli: 'v2.5.1',
-]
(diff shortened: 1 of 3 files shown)


vdice:
Tested 'live' to success w/ https://ci.deis.io/job/workflow-cli-release/5/console and https://ci.deis.io/job/workflow-cli-build-tag/3/console (edit: and also in https://ci.deis.io/job/workflow-cli-build-stable/1/console)

PR #237, opened 2016-09-23
fix(component_release_publish.groovy): add git publish step

vdice:
For pushing local commits to master branch of origin


vdice:
Tested/proven in https://ci.deis.io/job/component-release-publish/3/console

PR #238, opened 2016-09-23
docs(README.md): add flow diagrams

vdice:


jchauncey:
my only question is should we publish this some where in the docs?

vdice:
@jchauncey thanks for the reminder.  How about a link to the `When a Component is Tagged` section from the release docs?

If so, I'll issue the follow-up PR once this is merged.

mboersma (approved):
Excellent!

Issue #210, opened 2016-08-19
Notify slack via shell script/api curl


vdice:
Due to various issues (mostly mentioned in https://github.com/deis/jenkins-jobs/issues/192), we cannot use [our version](https://github.com/deis/jenkins-jobs/issues/153) of the Slack Plugin if we wish to dispatch to multiple slack channels _and_ keep this repo public.  As we place high priority on keeping this repo public, for our simple needs of notifying appropriate channels with information from job results, we propose at this time to build our own logic that we can use for this purpose and avoid the use of the Slack plugin altogether.

At a high level, this logic should be simple-as-possible and consist of, minimally, one script that curls slack’s [inbound webhook api](https://api.slack.com/incoming-webhooks) to send a given message to a given slack channel.  Confidential Slack tokens could be added to each repo in [repo.groovy](https://github.com/deis/jenkins-jobs/blob/master/repo.groovy) in the form of `channel` -> `credentials id of auth token`.

Once this logic exists in `bash/scripts`, any of our jobs would then be able to run the following (pseudo-code) from a `shell` step:

```
if <all good>; then 
  msg=<success msg>
else
  msg=<failure msg>
fi
slack-notify <component channel> <msg>
```

The work items of this ticket would then be:
1. the creation of the aforementioned logic, covering in minimal test
2. PoC use of aforementioned tool in a Jenkins job
3. If success, creation of appropriate ticket(s) to migrate jobs toward the use of this approach.

sgoings:
Why do we need a docker container to do this work if we're injecting secrets via Jenkins secrets + just curling slack's API? Just to handle the mapping?

vdice:
@sgoings great question -- indeed, it isn't crucial to encapsulate the logic as you mentioned in the form of a docker container.  

For an MVP of getting us off the slack notification plugin, we could just build this in the form of a bash script, just like the rest here in `bash/scripts` (plus get it under test) and add the necessary credentials binding value to each repo in the `repos` map in [repo.groovy](https://github.com/deis/jenkins-jobs/blob/master/repo.groovy) and we'd be then be able to invoke said functionality when needed, supplying the appropriate channel/msg args.

If the above sounds like a good/better idea, I can re-word this ticket or create a new issue for this and potentially close this.  Thoughts?

sgoings:
I'd definitely start with the simplest possible solution that could possibly work. No need to get all Docker-fancy before you need to.

vdice:
[...]

PR #243, opened 2016-09-26
ref(*): Notify Slack via api curl

vdice:
This refactors the component build and `workflow-test(-pr)` jobs to notify slack via its incoming webhook api, forgoing the use of the Slack Notifications plugin.  The big win here is the ability to send a message to a channel other than the global default channel (which the aforementioned plugin does not currently allow in a secure manner).

Therefore, the component build jobs can send job statuses to their appropriate channels and the downstream test jobs can report back to the same upstream component channels...

~~For this POC to be extended to jobs such as the `workflow-test(-pr)` jobs, we would
need to move the existing `repo.slackChannel` values from groovy and into the shell
script itself, so that it can be passed repo name values at runtime and dynamically
maps these to appropriate slack channel values.~~

TODO
- [x] remove `slack-notification-test-job`
- [ ] if proposal greenlit, follow-up PR to convert all jobs to using this approach

Ref https://github.com/deis/jenkins-jobs/issues/210
Closes https://github.com/deis/jenkins-jobs/issues/210

### DIFF
--- a/bash/scripts/slack_notify.sh
+++ b/bash/scripts/slack_notify.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+# slack-notify sends the provided message to the provided slack channel
+# using the incoming-webhook url expected to be defined in the job environment
+slack-notify() {
+  channel="${1}"
+  status="${2}"
+  message="${3}"
+
+  if [[ $# -lt 2 ]]; then
+    echo "Usage: slack-notify channel status [message]"
+    return 1
+  fi
+
+  title="${JOB_NAME} - ${BUILD_DISPLAY_NAME} *${status}* (<${BUILD_URL}|Open>)"
+
+  data='
+    {"channel":"'"${channel}"'",
+    "attachments": [
+      {
+        "fallback": "'"${title}"'",
+        "color": "'"$(get-status-color "${status}")"'",
+        "pretext": "'"${title}"'",
+        "text": "'"${message}"'",
+        "mrkdwn_in": ["pretext", "text"],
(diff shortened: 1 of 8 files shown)


bacongobbler:
If this works, I'm all for it. Custom messages on certain build failures would be great!

vdice:
As an example of the formatting foo/functionality that we would lose,

Formerly:

slack-plugin-example

Now:

slack-notify-example

That being said, the Incoming Webhooks API definitely supports gettin' more fancy: https://api.slack.com/docs/message-attachments

vdice:
@sgoings and/or @mboersma thoughts on this?  While we lose a bit of functionality and formatting foo, we now gain the ability to report job results (and other useful info) to channels particular to the component involved, clearing out the pile o noise from `#testing` -- or, rather, spreading that pile o noise around :)

mboersma:
> spreading that pile o' noise around

That seems like the goal--when everything goes into the #testing channel it's hard to sort the wheat from the chaff. In particular, a failure to promote an artifact after merging to master should be targeted to an appropriate notification channel so it is more visible.

sgoings:
@vdice if you could add some color to those messages, that'd be 💯

bacongobbler:
+1 on the pretty colors notion

vdice:
@sgoings @bacongobbler here is the current formatting from this PR.  Thoughts/suggestions?

notification-formatting-updated

bacongobbler:
LGTM! :shipit:

vdice:
I'll build all jobs off of this branch so we can 'demo' this change for a spell before merging.

Issue #174, opened 2016-08-01
Component release e2e strategy


vdice:
https://github.com/deis/jenkins-jobs/issues/158 rolled out the initial iteration of a component release pipeline.  The e2e stage of this pipeline simply retrieved the last released `workflow-<release>` chart, plugged in the release candidate for its respective component docker tag and ran that against the `workflow-<release>-e2e` chart.

The drawback to this approach is other components may have released in the interim since the last released workflow chart.  For the most accurate test results, we would want an up-to-date `workflow` chart wherein all components' docker tags are the very latest semver tags straight from git or image registry.  

To save manual effort from having to look up each component's latest semver tags, a corresponding tool would need to be made (perhaps in https://github.com/deis/deisrel) that could ping either the GitHub API or Quay.io/DockerHub API to derive each component's latest semver tag and produce the chart used for testing...

Joshua-Anderson:
It would be pretty trivial to add an option to `deisrel` that after looking up tag versions would edit the `generate_params.toml` file with the latest versions. Should I implement that?

vdice:
@Joshua-Anderson Indeed, I think that would be a great option for us, at least for the API/`generate_params.toml` update portion.  Thinking on it more, our current e2e infra (specifically the logic in https://github.com/deis/e2e-runner) doesn't yet really support plugging in a chart on-demand.  (Rather it fetches and installs charts from a git remote.)  So, I think there may be another sibling chunk of work to enable 'passing' a chart to install to e2e-runner... unless there might be a better way?

PR #245, opened 2016-09-27
ref(component_release.groovy): remove e2e stage from component release pipeline

vdice:
Here I propose to remove the e2e stage from the component release [pipeline](https://github.com/deis/jenkins-jobs#when-a-component-is-tagged) for a few reasons.

First, the way it is set up now is less than ideal -- see [here](https://github.com/deis/jenkins-jobs/issues/174) for a bit of background.  Long story short, we are testing a component release candidate against what is most likely an outdated release chart.

Secondly, this step can easily be considered overkill.  In tagging a component for release, it is assumed that one is tagging a commit in master that has already been vetted by e2e (see existing master merge [pipeline](https://github.com/deis/jenkins-jobs/#when-a-component-pr-is-merged-to-master)).  

Thirdly, as we've started to release components individually, we are seeing the need to unshackle each component from having to run the full Workflow chart e2e against every change (this process being very expensive and oftentimes fraught with flaky/indeterminant failures), instead opting to strengthen the component-centric test suites.  (Note that this stage still exists in component PR and Master pipelines.)


helgi:
shipt it

mboersma:
I agree: running e2e all over again at that point is overkill, and may not be meaningful in the context of the last release chart anyway. I favor of running all the tests, all the time (if possible), but in this case less is more.

PR #246, opened 2016-09-27
feat(workflow_cli_release.groovy): cgo-enabled darwin binary support

vdice:
Sets up the workflow-cli release pipeline to promote official release `darwin-amd64` binaries that have been built on an actual OSX slave using `CGO_ENABLED=1`.

~~Requires https://github.com/deis/workflow-cli/pull/252~~

Fixes https://github.com/deis/workflow-cli/issues/246

TODO
- [x] remove test bucket override

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,34 @@ Publish───────────────────────┐
 └─────────────────────────────┘
 ```
 
+### When Workflow-CLI is tagged
+```
+Trigger─────────────────────────────────┐
+│                                       │   - triggered by `v1.2.3` git tag push webhook
+│       "workflow-cli-release"          │   - pass this to downstream job(s)
+│                                       │
+└──────────┬────────────────────────────┘
+           │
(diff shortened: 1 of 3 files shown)


vdice:
Having trouble setting up the go environment on the mac/darwin slave correctly in this updated job and hoping there may be an easier/simpler solution... removing the 'LGTM' for now and will update in time.

vdice:
@gerred @lachie83 when you have a moment, could you test that the darwin amd64 binaries produced by the jobs PRed here fix the original issue (https://github.com/deis/workflow-cli/issues/246)?  

```
curl https://storage.googleapis.com/workflow-cli-test-bucket/v2.6.1/deis-v2.6.1-darwin-amd64 -o deis && chmod +x deis
```

robholland:
That binary works for us :)

gerred:
Works for me! :+1:

PR #254, opened 2016-10-07
fix(component_release_publish): disable concurrent builds

vdice:
Fix for the following publish job failure: https://ci.deis.io/job/component-release-publish/32/console


jchauncey:
g2g

PR #255, opened 2016-10-07
fix(workflow_test.groovy): don't attempt to update commit status

vdice:
if ACTUAL_COMMIT is not defined/present.

Better fix to related/flawed fix in https://github.com/deis/jenkins-jobs/pull/250

cc @jchauncey


jchauncey:
g2g

PR #266, opened 2016-10-15
Refactor website jobs

slack:
- Rename `deploy_website` to `deis-com-deploy` and update references.
- Add `deis-io-deploy` to deploy deis.io to prod/staging s3+cloudfront


slack:
Requires https://github.com/deis/deis.io/pull/186

PR #267, opened 2016-10-17
chore(tests): build pr and merges

slack:
Refs https://github.com/deis/deis.com/issues/285


slack:
Requires https://github.com/deis/deis.io/pull/187

PR #269, opened 2016-10-19
feat(repo): add jenkins-node to CI

bacongobbler:
See https://github.com/deis/jenkins-node

### DIFF
--- a/repo.groovy
+++ b/repo.groovy
@@ -30,6 +30,11 @@ repos = [
     workflowComponent: true,
     chart: 'fluentd'],
 
+  [ name: 'jenkins-node',
+    components: [[name: 'jenkins-node']],
+    slackChannel: '#testing',
+    chart: 'jenkins-node'],
+
   [ name: 'logger',
     components: [[name: 'logger']],
     slackChannel: '#logger',


repo.groovy
```diff
@@ -30,6 +30,11 @@ repos = [
     workflowComponent: true,
     chart: 'fluentd'],
 
+  [ name: 'jenkins-node',
```
vdice, line 33: @bacongobbler do we also want the usual component build and release jobs to be auto-generated for `jenkins-node` (really just necessitates that the project have a `Makefile` w/ `docker-build` and `docker-push` targets)?  

If no, we can add an add'l key:value of `buildJobs: false` in this same entry -- then the only job that will be generated will be `jenkins-node-chart-publish` -- which is currently usually kicked off as part of a component's release pipeline (which we wouldn't have w/ `buildJobs: false` mentioned above), but can be kicked off manually as well.

bacongobbler: Certainly! There's no Makefile in the project at the moment but I can cook one up in a sec. :)

bacongobbler: Done. Is there anything else that I need to do here?

vdice: @bacongobbler Nope, _should_ be g2g!

PR #270, opened 2016-10-19
bug(merge builds): scm ref needs to be concrete

slack:


slack:
A bit too much faffing about on my part. 🗡

sgoings:
Why the change?

slack:
From what I could tell the literal string of `${DEIS_IO_ORG}/${DEIS_IO_REPO}` was showing up as the repo url... So the hooks to watch the repo/branch aren't added properly.

PR #271, opened 2016-10-20
fix(helm): Update the helm to beta1 and change the repo index command

kmala:


vdice:
@kmala now that I see how many places we need to change, do we want to put the default helm version in one location?  We could create a `version: v2.0.0-beta.1` key/value in the map [here](https://github.com/deis/jenkins-jobs/blob/master/common.groovy#L35) and then all jobs can switch to just using `defaults.helm.version` for their parameters. (And the version can hence be updated in the one spot)

vdice (approved):
Looks great

PR #273, opened 2016-10-20
feat(charts): Upload the values file separately so it can be fetched easily before install

kmala:
This would be useful during install process which removes the effort of the user to fetch and untar the chart before installing


vdice:
@kmala LGTM -- when convenient, can you reference the corresponding docs PR(s)?

kmala:
this is the docs PR https://github.com/deis/workflow/pull/564

PR #274, opened 2016-10-20
fix(charts):Use proper name of the repo to fetch the latest tag from wfm api

kmala:


jobs/workflow_chart_pipeline.groovy
```diff
@@ -78,9 +78,10 @@ job("${chartRepo.staging}-chart-publish") {
 
           ## update requirements.yaml with latest tags for each component
           for component in ${components}; do
-            latest_tag="\$(get-latest-component-release "\${component}")"
-            perl -i -0pe "s/<\${component}-tag>/\${latest_tag}/g" ${chart}/requirements.yaml
-            helm repo add "\${component}" "https://charts.deis.com/\${component}"
+            IFS=':' read -r -a repo <<< "\${component}"
```
vdice: I'd suggest changing `repo` to something slightly more indicative of what it actually represents... perhaps something like `chartAndRepo`.  

It might also be a readability improvement if appropriate vars are created for use in the next few lines:

```
chart="\${chartAndRepo[0]}"
repo="\${chartAndRepo[1]}"
```

PR #282, opened 2016-11-07
feat(*): add scripts/jobs for signing helm charts

vdice:
ref https://github.com/deis/workflow/issues/595

TODO
- [x] test affected pipelines 
  * workflow: https://ci.deis.io/job/workflow-chart-publish/
  * component: https://ci.deis.io/job/router-chart-publish/ (just one component needed)

### DIFF
--- a/jobs/component_chart_publish.groovy
+++ b/jobs/component_chart_publish.groovy
@@ -59,45 +59,66 @@ repos.each { Map repo ->
       }
 
       steps {
-        shell """
-          #!/usr/bin/env bash
+        shell new File("${workspace}/bash/scripts/helm_chart_actions.sh").text +
+          """
+            #!/usr/bin/env bash
 
-          set -eo pipefail
-          set -x
+            set -eo pipefail
+            set -x
 
-          # check out RELEASE_TAG tag
-          commit="\$(git checkout "\${RELEASE_TAG}" && git rev-parse HEAD)"
+            # check out RELEASE_TAG tag
+            commit="\$(git checkout "\${RELEASE_TAG}" && git rev-parse HEAD)"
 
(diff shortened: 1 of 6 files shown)


jobs/component_chart_publish.groovy
```diff
@@ -59,45 +59,46 @@ repos.each { Map repo ->
...
+            #!/usr/bin/env bash
+
+            set -eo pipefail
+            set -x
+
+            # check out RELEASE_TAG tag
+            commit="\$(git checkout "\${RELEASE_TAG}" && git rev-parse HEAD)"
+
+            if [ -d charts ]; then
+              cd charts
```
bacongobbler, line 73: do we want to `pushd/popd` from this directory so the next script (if we include another at some point) is running from the original directory, or is this fine the way it is?

vdice: @bacongobbler I actually haven't used those utilities before, but if it cleans up and/or makes these script(s) less error-prone, I'd definitely endorse it.

vdice:
realize I also need to hook up the existing component/workflow chart pipelines to this new `helm-chart-sign` job... additions coming soon

vdice:
additions mentioned in https://github.com/deis/jenkins-jobs/pull/282#issuecomment-258993169 added; next need to test affected jobs by building them off this branch... **edit: done**

jobs/component_chart_publish.groovy
mboersma: It doesn't look like this script does a `cd ..` later, so probably `pushd/popd` aren't necessary here.
https://en.wikipedia.org/wiki/Pushd_and_popd

PR #285, opened 2016-11-10
feat(jobs/helm_chart_sign.groovy): add helm chart verify job

vdice:
Adds a final verification step for verifying a signed chart.


vdice:
Tested to success in https://ci.deis.io/job/helm-chart-verify/2/console

PR #289, opened 2016-11-15
fix(jobs): upload an index file if it does not exist

bacongobbler:
When creating a release for the first time, an index file is not present in the bucket to copy and merge from. Therefore in this case we just want to upload an initial index file and call it a day.

Note that this will not help jenkins-node, but will prevent this issue from occurring in the future.

### DIFF
--- a/jobs/component_chart_publish.groovy
+++ b/jobs/component_chart_publish.groovy
@@ -86,11 +86,16 @@ repos.each { Map repo ->
               # package release chart
               helm package ${repo.chart}
 
-              # download index file from aws s3 bucket
-              aws s3 cp s3://helm-charts/${repo.chart}/index.yaml .
+              if [ -z "\$(aws s3 ls s3://helm-charts/${repo.chart}/index.yaml)" ]; then
+                # the index file does not exist yet, so let's create it
+                helm repo index . --url https://charts.deis.com/${repo.chart}
+              else
+                # download index file from aws s3 bucket
+                aws s3 cp s3://helm-charts/${repo.chart}/index.yaml .
 
-              # update index file
-              helm repo index . --url https://charts.deis.com/${repo.chart} --merge ./index.yaml
+                # update index file
+                helm repo index . --url https://charts.deis.com/${repo.chart} --merge ./index.yaml
+              fi
 
               # push packaged chart and updated index file to aws s3 bucket
               aws s3 cp ${repo.chart}-\${RELEASE_TAG}.tgz s3://helm-charts/${repo.chart}/ \


jobs/component_chart_publish.groovy
```diff
@@ -86,11 +86,16 @@ repos.each { Map repo ->
               # package release chart
               helm package ${repo.chart}
 
-              # download index file from aws s3 bucket
-              aws s3 cp s3://helm-charts/${repo.chart}/index.yaml .
-
-              # update index file
-              helm repo index . --url https://charts.deis.com/${repo.chart} --merge ./index.yaml
+              if [ "$(git tag | wc -l)" == "1" ]; then
```
vdice: we need to escape the `$` here (`\$`) b/c it is a shell/runtime resolution as opposed to a groovy/job build-time resolution...

vdice (changes requested):

vdice:
minor change needed; otherwise LGTM!

bacongobbler:
minor change addressed

vdice (approved):

kmala (approved):

bacongobbler:
The code has been changed to accomodate if the file does not exist, please take a look before this is merged.

The code was thanks to http://stackoverflow.com/a/27003199.

bacongobbler:
Hold up on this PR. Just a thought, since this job runs when no index file exists, should we run `helm package` initially, upload the index file then re-run `helm package` and merge it with the now-present index? Just wondering if uploading a bare index file when we're releasing a tag of a chart is what we intend.

jobs/component_chart_publish.groovy
```diff
@@ -86,11 +86,16 @@ repos.each { Map repo ->
               # package release chart
               helm package ${repo.chart}
 
-              # download index file from aws s3 bucket
-              aws s3 cp s3://helm-charts/${repo.chart}/index.yaml .
-
-              # update index file
-              helm repo index . --url https://charts.deis.com/${repo.chart} --merge ./index.yaml
+              if [ "\$(aws s3 ls s3://helm-charts/${repo_chart}/index.yaml | wc -l)" == "0" ]; then
```
vdice: could also use `if [ -z "\$(aws s3 ls s3://helm-charts/${repo_chart}/index.yaml)" ]; then` here -- but just an fyi

vdice:
> should we run `helm package` initially

@bacongobbler we do a helm package immediately prior, so the first `index.yaml` file _should_ actually be populated... so I think we are good here.  But good concern.

jobs/component_chart_publish.groovy
bacongobbler: done

jobs/component_chart_publish.groovy
```diff
@@ -86,11 +86,16 @@ repos.each { Map repo ->
               # package release chart
               helm package ${repo.chart}
 
-              # download index file from aws s3 bucket
-              aws s3 cp s3://helm-charts/${repo.chart}/index.yaml .
-
-              # update index file
-              helm repo index . --url https://charts.deis.com/${repo.chart} --merge ./index.yaml
+              if [ -z "\$(aws s3 ls s3://helm-charts/${repo_chart}/index.yaml)" ]; then
```
vdice: ah, also `s/repo_chart/repo.chart`

Issue #286, opened 2016-11-10
[Meta] Transition Workflow and component pipelines to supporting use of Kubernetes Helm charts


vdice:
Workflow and component pipelines (and various other moving parts) need modifications to support the use of Kubernetes Helm charts.  The long-term goal is to remove support for the Helm Classic charts altogether, as we transition to the official Helm.

### e2e-runner updates: 

- [x] Update [e2e-runner](https://github.com/deis/e2e-runner) logic to run `--set <component>.docker_tag=git-<short sha for commit>` for any/every `<COMPONENT>_GIT_TAG` set in the environment.  (See [here](https://github.com/deis/e2e-runner/blob/master/scripts/functions.sh) for context around the logic setting the vars) (https://github.com/deis/e2e-runner/pull/61)

- [x] Update logic [here](https://github.com/deis/e2e-runner/blob/master/scripts/run.sh#L31) so to remove hard-coded `workflow-dev` chart repo in favor of a configurable env var (say, `$CHART_REPO`) so that it may be `workflow-pr`, `workflow-staging`, etc. (https://github.com/deis/e2e-runner/pull/60)

- [x] ~~Add `WORKFLOW` to this [list](https://github.com/deis/e2e-runner/blob/master/scripts/config.sh#L38) so that a `WORKFLOW_SHA` env var can be detected and used (see below in deis/workflow PR case)~~ Not needed
 
### Workflow pipeline updates:
- [x] Update the [workflow chart pipeline](https://github.com/deis/jenkins-jobs/blob/master/jobs/workflow_chart_pipeline.groovy) with the following:
    - [x] Prefix the latest tag for each component with `>=` to enable matching within a range, if repo is `workflow-dev` (Note: depends on https://github.com/kubernetes/helm/issues/1351) (https://github.com/deis/jenkins-jobs/pull/288)
  * Is there a change to the `charts` subdirectory?  If Yes:
    - [x] PR commit?: (https://github.com/deis/jenkins-jobs/pull/294)
      * Package and push `workflow-<version>+<git sha>.tgz` tarfile to the `workflow-pr` chart repo/bucket
      * kick off downstream e2e job specifying `CHART_REPO_TYPE=pr` and `WORKFLOW_SHA=<git sha>`
    - [x] Master commit: (https://github.com/deis/jenkins-jobs/pull/288)
      * Package and push `workflow-<version>-<timestamp>+<git sha>.tgz` tarfile to the `workflow-dev` chart repo/bucket
[...]

PR #294, opened 2016-11-22
Add k helm pr support

vdice:
Closes https://github.com/deis/jenkins-jobs/issues/286

TODO:
 - [x] test affected pipelines on Jenkins
 - [x] create follow-up ticket removing use of older `workflow-test(-pr)` jobs in favor of the newer e2e job in `workflow_chart_pipeline.groovy` (and/or merging the two) (https://github.com/deis/jenkins-jobs/issues/299)

### DIFF
--- a/bash/scripts/publish_helm_chart.sh
+++ b/bash/scripts/publish_helm_chart.sh
@@ -11,9 +11,18 @@ publish-helm-chart() {
   local chart="${1}"
   local repo_type="${2}"
 
-  # check out RELEASE_TAG tag (if empty, just stays on master commit)
-  short_sha="${SHORT_SHA:-$(git checkout -q "${RELEASE_TAG}" && git rev-parse --short HEAD)}"
-  git_tag="${GIT_TAG:-$(git describe --abbrev=0 --tags)}"
+  # give ACTUAL_COMMIT precedence
+  if [ -n "${ACTUAL_COMMIT}" ]; then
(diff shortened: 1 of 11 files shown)


bash/scripts/publish_helm_chart.sh
```diff
@@ -79,16 +90,27 @@ update-chart() {
...
         component_chart_repo="${component_chart}-dev"
       fi
+
       perl -i -0pe 's/<'"${component_chart}"'-tag>/"'"${component_chart_version}"'"/g' "${chart}"/requirements.yaml
       perl -i -0pe 's='"${DEIS_CHARTS_BASE_URL}/${component_chart}\n"'='"${DEIS_CHARTS_BASE_URL}/${component_chart_repo}\n"'=g' "${chart}"/requirements.yaml
       helm repo add "${component_chart_repo}" "${DEIS_CHARTS_BASE_URL}/${component_chart_repo}"
     done
 
+    # DEBUG
+    cat "${chart}"/requirements.yaml
```
bacongobbler, line 122: is this no longer required? We should remove it if so

vdice: @bacongobbler alas, this continues to help us investigate/troubleshoot failed Workflow chart packaging... I think we want to keep this `cat` for the time being.  Therefore, I'll remove the `# DEBUG` comment above.

PR #301, opened 2016-12-08
feat(repo.groovy): add imageBuild bool

vdice:
Specifically for designating if a Workflow component does not require
an image build/test pipeline but _does_ require a component chart pipeline,
as is the case in https://github.com/deis/registry-proxy/pull/9

### DIFF
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -134,13 +134,17 @@ repos.each { Map repo ->
 
               git_commit="\$(get-actual-commit ${repo.name} \${ghprbActualCommit})"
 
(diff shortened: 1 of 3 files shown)


bacongobbler:
While this isn't necessary for registry-proxy any more, I still think this can be useful for meta-charts (e.g. if deis/workflow wasn't dual-purpose as both a component serving the docs and as a meta chart, but JUST as a chart alone).

vdice:
an alternative route was pursued for registry-proxy (https://github.com/deis/workflow/issues/644) si this is no longer needed; closing.

PR #302, opened 2016-12-08
chore(repo): remove registry-proxy from CI

bacongobbler:
See https://github.com/deis/workflow/issues/644


vdice:
Note that we'll need to manually delete all `registry-proxy*` jobs in ci.deis.io after this is merged to 'fully clean up'

bacongobbler:
I will do that^

PR #307, opened 2016-12-15
fix(deis_com_deploy): build deis.com on a pushed tag of deis/workflow

bacongobbler:
here's my attempt at fixing an issue when going through a patch release with https://github.com/deis/workflow/pull/662

### DIFF
--- a/jobs/deis_com_deploy.groovy
+++ b/jobs/deis_com_deploy.groovy
@@ -31,6 +31,7 @@ job(name) {
          relativeTargetDirectory('workflow')
      }
      branch('master')
+     branch('*/tags/*')
    }
    git {
      remote {


vdice:
I _think_ this _should_ work... to be sure, we could either build this altered job off of this branch in Jenkins or create a similar test job w/ both branches and push a git tag (perhaps watching one of our forked repos, etc.)...

bacongobbler:
Wish there was docs for this in the ghprb plugin! This is a shot in the dark, just verified it was correct syntax in the DSL syntax checker.

I'd like to try building this altered job in Jenkins. I'd be happy to try it out and see if it works. Worst case scenario, it either

a) fails to build the job
b) the job now builds from only tags (read: plenty of time to revert)

So to do that, I'd just run [seed-job](https://ci.deis.io/job/seed-job/) and reference this branch? It doesn't look like there's any configurable parameters in that job so it only builds off what is in master.

We could just merge this and revert if things fail. 👿

bacongobbler:
This worked nicely: https://ci.deis.io/job/deis-com-deploy/configure

Turns out you can specify any number of branch specifiers. How neat is that? 🐛

PR #315, opened 2017-01-03
fix(component_chart_publish.groovy): fix dsl

vdice:
(missing `steps` block)


vdice:
tested via https://ci.deis.io/job/router-chart-publish/29/console

Issue #293, opened 2016-11-21
remove helmc support


vdice:
To remove helmc support from CI, delete all code marked with `helmc-remove` (marks added in https://github.com/deis/jenkins-jobs/pull/292)

Tentatively scheduling for the v2.11 milestone...

bacongobbler:
assigning this to you @vdice but please feel free to re-assign if you can't get to it this release.

PR #323, opened 2017-01-11
ref(*): remove all helmc logic

vdice:
Closes https://github.com/deis/jenkins-jobs/issues/293


mboersma:
`+0 -413 LOC` is always a good thing.

PR #325, opened 2017-01-12
ref(*): add e2e-runner to repo list

vdice:
So e2e-runner PRs/master merges processed as other Workflow & co. repos,
instead of maintaining separate job (highly similar and prone to diverging)

TODO
 - [x] Test in Jenkins by running PR and master e2e-runner pipelines

PR #327, opened 2017-01-20
fix(component_jobs): pass "--no-cache" to docker builds

mboersma:
Sets an env var that other components (will) respect, which should avoid docker cached layers in CI.

### DIFF
--- a/jobs/component_jobs.groovy
+++ b/jobs/component_jobs.groovy
@@ -137,7 +137,8 @@ repos.each { Map repo ->
 
               export IMAGE_PREFIX=deisci VERSION="git-\${git_commit:0:7}"
               docker login -e="\$DOCKER_EMAIL" -u="\$DOCKER_USERNAME" -p="\$DOCKER_PASSWORD"
-              DEIS_REGISTRY='' make docker-build ${dockerPush}
+              # build once with "docker --pull --no-cache" to avoid stale layers
+              DEIS_REGISTRY='' DOCKER_BUILD_FLAGS="--pull --no-cache" make docker-build ${dockerPush}
               docker login -e="\$QUAY_EMAIL" -u="\$QUAY_USERNAME" -p="\$QUAY_PASSWORD" quay.io
               DEIS_REGISTRY=quay.io/ make docker-build ${dockerPush}
 


jobs/component_jobs.groovy
```diff
@@ -137,7 +137,8 @@ repos.each { Map repo ->
 
               export IMAGE_PREFIX=deisci VERSION="git-\${git_commit:0:7}"
               docker login -e="\$DOCKER_EMAIL" -u="\$DOCKER_USERNAME" -p="\$DOCKER_PASSWORD"
-              DEIS_REGISTRY='' make docker-build ${dockerPush}
+              # build once with "docker --pull --no-cache" to avoid stale layers
```
vdice, line 140: perhaps we can just `export DOCKER_BUILD_FLAGS="--pull --no-cache"` above (like IMAGE_PREFIX and VERSION) so that both `docker-build` invocations use it (assuming this is desired) -- note the `quay.io` docker-build below...

jobs/e2e_runner.groovy
```diff
@@ -98,7 +98,8 @@ evaluate(new File("${workspace}/common.groovy"))
         make docker-test
 
         docker login -e="\$DOCKER_EMAIL" -u="\$DOCKER_USERNAME" -p="\$DOCKER_PASSWORD"
-        DEIS_REGISTRY='' make docker-build ${dockerPush}
+        # build once with "docker --pull --no-cache" to avoid stale layers
```
vdice: This won't have to be changed if/when https://github.com/deis/jenkins-jobs/pull/325 is merged; though no biggie to resolve merge conflict depending on order...

vdice (changes requested):

jobs/component_jobs.groovy
mboersma: We could--but then when we build once for Docker Hub and once for quay.io, the second build *won't* use cached layers and will `apt-get` and such all over again. That's slower and introduces the potential for them to disagree.

I think we want the two images to contain the same bits instead, so this is one way of doing that.

jobs/e2e_runner.groovy
mboersma: No worries--if that gets merged first (nudge @bacongobbler), I'll clean this up.

jobs/component_jobs.groovy
vdice: Ah good point -- totes makes sense! 👍

vdice (approved):

jobs/e2e_runner.groovy
bacongobbler: thanks for the nudge. :)

PR #328, opened 2017-01-20
feat(jobs): add workflow upgrade test job

vdice:
Ref/Requires https://github.com/deis/e2e-runner/pull/80


vdice:
Tested in https://ci.deis.io/job/workflow-upgrade-test/4/console

PR #332, opened 2017-02-15
chore(common.groovy): bump helm to v2.2.0

vdice:


vdice:
no rush on this; there *may* be some issues in current 2.2.0 binaries regarding HELM_HOME... would be a good idea to build jenkins off this branch and run some chart assembly pipelines unless it turns out to be a non-issue.

bacongobbler:
code LGTM either way.

mboersma:
[Helm v2.2.1](https://github.com/kubernetes/helm/releases/tag/v2.2.1) is out now, which fixes the $HOME_HOME-related issue. We should use that instead.

PR #337, opened 2017-03-08
Revert "chore(helm_chart_actions.sh): update registry-proxy logic"

vdice:
Reverts deis/jenkins-jobs#331

ref https://github.com/deis/workflow/pull/765


bacongobbler:
I was just about to suggest that this be reverted as well. LGTM!

PR #338, opened 2017-03-09
fix(component_jobs.groovy): use 'dev' repo for non-chart PRs

vdice:
So that downstream test job does not install Workflow from 'pr' repo,
which may be a broken chart.


mboersma (approved):

vdice:
Closing in favor of https://github.com/deis/e2e-runner/pull/89

PR #350, opened 2017-04-05
fix(workflow_chart_pipeline.groovy): remove git checkout cmd

vdice:
tested via https://ci.deis.io/job/workflow-chart-release/8/console

### DIFF
--- a/jobs/workflow_chart_pipeline.groovy
+++ b/jobs/workflow_chart_pipeline.groovy
@@ -410,8 +410,6 @@ job("${chart}-chart-release") {
 
         download-and-init-helm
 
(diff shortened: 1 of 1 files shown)


mboersma (approved):

bacongobbler (approved):

bacongobbler:
Just background context, what broke due to this or what are we fixing? Just that helm no longer requires you to check out release tags of the source code?

vdice:
@bacongobbler good question -- this is fixing a regression from an earlier refactor wherein this line was mistakenly added.  We have no need to check out the tag as we're just fetching and updating resources from the chart repo using `helm` and `az`.

PR #356, opened 2017-04-10
ref(jobs/component_chart_publish.groovy): update job

vdice:
Add KEYRING param specifying exact keyring location
Only kick off workflow-chart-publish if repo is Workflow component


vdice:
Closing in favor of a different fix (updating volume mount for our deis-signatory container to use the new JENKINS_HOME value of `/var/jenkins_home`).  We'll be re-vamping our signatory node approach in the near future as well, so this should be fine for now.

PR #369, opened 2017-04-13
feat(repo.groovy): Remove influxdb from component list

jchauncey:
Because we are no longer actively maintaining an influxdb setup for the monitoring stack we should not have any jobs run for it.

depends on https://github.com/deis/monitor/pull/185


jchauncey:
We are not shipping the stable/influx chart yet so we can hold off on this.

PR #372, opened 2017-04-21
fix(jobs/component_release.groovy): check if tag already released

vdice:
Jenkins has a bizarre quirk of sometimes attempting to release a given tag multiple times, having been incorrectly triggered by other push events.  Until we figure out the root cause, this quick check avoids attempting an subsequent releases of a tag.

### DIFF
--- a/bash/scripts/get_latest_tag.sh
+++ b/bash/scripts/get_latest_tag.sh
@@ -3,6 +3,8 @@
 set -eo pipefail
 
 get-latest-tag() {
+  component="${1}"
+
   tag=''
   if [ -n "${TAG}" ]; then
     tag="${TAG}"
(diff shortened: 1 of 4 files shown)


vdice:
tested via https://ci.deis.io/job/jenkins-release/8/

mboersma (approved):

bash/scripts/get_latest_tag.sh
```diff
@@ -11,5 +13,12 @@ get-latest-tag() {
   else
     exit 1
   fi
+
+  # If tag already released, bail out
+  if curl -sf "https://versions.deis.com/v3/versions/stable/deis-${component}/${tag}"; then
```
bacongobbler: The only thing I'm concerned about is that this will exit 1 if versions.deis.com is down or unresponsive with no indication that it failed because it was down. Is there a way we can report the error that is returned here so we can determine next steps to remediate the issue?

bacongobbler (approved):
Overall I approve though and we can fix this to get this unblocked

bash/scripts/get_latest_tag.sh
vdice: @bacongobbler that's definitely a valid concern.  How about I remove the `s` (silent) flag and add the `v` (verbose) flag in its stead so we get the full output of why the curl command might have failed?

bacongobbler: I think there's a flag that'll show the error when it actually fails. `-S`? So `curl -sSf`

PR #380, opened 2017-05-16
feat(jobs/*): Adds cloud provider env var to e2e jobs

jchauncey:
This will be defaulted to google for now

### DIFF
--- a/common.groovy
+++ b/common.groovy
@@ -34,6 +34,7 @@ defaults = [
     storageAccountKeyID: 'c9cad0b0-53dd-4d38-a832-c4b29aeaf49b',
   ],
   statusesToNotify: ['SUCCESS', 'FAILURE'],
+  "e2eRunner": [ provider: 'azure', ],
 ]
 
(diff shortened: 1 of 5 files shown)


arschles:
@jchauncey LGTM. this seems to require https://github.com/deis/k8s-claimer/pull/104 to be merged & deployed, right?

common.groovy
```diff
@@ -34,6 +34,7 @@ defaults = [
     storageAccountKeyID: 'c9cad0b0-53dd-4d38-a832-c4b29aeaf49b',
   ],
   statusesToNotify: ['SUCCESS', 'FAILURE'],
+  "e2eRunner": [ provider: 'google', ],
```
sgoings: We should default to azure I think.

arschles:
Still LGTM after your recent changes @jchauncey

common.groovy
jchauncey: updated

jchauncey:
reran this job and it succeeded so im merging it - https://ci.deis.io/job/jenkins-jobs-pr/95/console

PR #381, opened 2017-05-17
feat(k8s-claimer): Use charts to deploy instead of workflow

jchauncey:
This PR does a number of things

* Use the k8sclaimer chart to deploy instead of workflow
* Creates a master build cli job and uploads to azure blob storage
* Creates a PR job for running tests, building the cli, and building the binary/image. It will not upload the artifacts or the image to quay!

### DIFF
--- a/jobs/k8s_claimer.groovy
+++ b/jobs/k8s_claimer.groovy
@@ -4,6 +4,116 @@ evaluate(new File("${workspace}/common.groovy"))
 
 slackChannel = '#ops'
 
+job('k8s-claimer-pr') {
+  description """
+  <p>Run the tests for k8s-claimer</p>
+  <p>
+    K8s-Claimer serves as a Kubernetes cluster leaser for running Workflow E2E tests in CI.
+  </p>
+  """.stripIndent().trim()
+
+  scm {
+    git {
+      remote {
+        github("deis/k8s-claimer")
+        credentials('597819a0-b0b9-4974-a79b-3a5c2322606d')
+        refspec('+refs/pull/*:refs/remotes/origin/pr/*')
+      }
+      branch('${sha1}')
+    }
+  }
+
+  def statusesToNotify = ['FAILURE']
+  publishers {
+    postBuildScripts {
+      onlyIfBuildSucceeds(false)
+      steps {
+        statusesToNotify.each { buildStatus ->
+          conditionalSteps {
+            condition {
+              status(buildStatus, buildStatus)
+              steps {
+                shell new File("${workspace}/bash/scripts/slack_notify.sh").text +
+                  """
+                    slack-notify "${slackChannel}" "${buildStatus}"
+                  """.stripIndent().trim()
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  concurrentBuild()
+  throttleConcurrentBuilds {
+    maxPerNode(defaults.maxBuildsPerNode)
+    maxTotal(defaults.maxTotalConcurrentBuilds)
+  }
+
+  logRotator {
+    daysToKeep defaults.daysToKeep
+  }
+
+  triggers {
+    pullRequest {
+      admin('deis-admin')
+      cron('H/5 * * * *')
+      useGitHubHooks()
+      triggerPhrase('OK to test')
+      orgWhitelist(['deis'])
+      allowMembersOfWhitelistedOrgsAsAdmin()
+      // this plugin will update PR status no matter what,
+      // so until we fix this, here are our default messages:
+      extensions {
+        commitStatus {
+          context('ci/jenkins/pr')
+          triggeredStatus("Triggering k8s-claimer build/test pipeline...")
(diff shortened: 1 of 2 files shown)


jobs/k8s_claimer_deploy.groovy
```diff
@@ -60,29 +60,46 @@ job(name) {
...
     shell '''
       #!/usr/bin/env bash
-
       set -eo pipefail
+      echo $KUBECONFIG_BASE64 | base64 --decode > kubeconfig
+
+      curl -Ls -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v$K8S_VERSION/bin/linux/amd64/kubectl
+	    chmod +x /usr/bin/kubectl
+
+      curl -f --silent --show-error --retry 5 --retry-delay 10 -O https://storage.googleapis.com/kubernetes-helm/helm-v2.4.1-linux-amd64.tar.gz
```
vdice: let's use an env var for `HELM_VERSION` instead of hard-coded v2.4.1 (common.groovy should define `defaults.helm.version`, I believe)

jobs/k8s_claimer_deploy.groovy
```diff
@@ -60,29 +60,46 @@ job(name) {
...
   }
 
   steps {
     shell '''
       #!/usr/bin/env bash
-
       set -eo pipefail
+      echo $KUBECONFIG_BASE64 | base64 --decode > kubeconfig
+
+      curl -Ls -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v$K8S_VERSION/bin/linux/amd64/kubectl
```
vdice: looks like we still need to add `K8S_VERSION` to job env for this to resolve?

jchauncey: Good catch. I meant to just hardcode it for now

jobs/k8s_claimer_deploy.groovy
jchauncey: updated to just use the defaults.helm.version

jobs/k8s_claimer_deploy.groovy
jchauncey: i removed this because i wont need it

jobs/k8s_claimer_deploy.groovy
```diff
@@ -60,29 +60,43 @@ job(name) {
...
   }
 
   steps {
     shell '''
       #!/usr/bin/env bash
-
       set -eo pipefail
+      echo $KUBECONFIG_BASE64 | base64 --decode > kubeconfig
+
+      curl -f --silent --show-error --retry 5 --retry-delay 10 -O https://storage.googleapis.com/kubernetes-helm/helm-${defaults.helm.version}-linux-amd64.tar.gz
```
vdice: if/when we want to interpolate groovy-land vars (like here w/ `defaults.helm.version`) we'll need to change the wrapping `shell` block to use double quotes instead of literal/single quotes (line 78 as of writing)

Note that once we do that, we need to escape all shell-land vars, e.g. `${PATH}` becomes `\${PATH}` so as to prevent groovy interpolation.

jchauncey: updated PR to fix this

jobs/k8s_claimer_deploy.groovy
```diff
@@ -60,29 +60,43 @@ job(name) {
...
     }
   }
 
   steps {
-    shell '''
+    shell """
       #!/usr/bin/env bash
-
       set -eo pipefail
+      echo $KUBECONFIG_BASE64 | base64 --decode > kubeconfig
```
vdice: looks like this one needs escaping here, too (conveniently, the dsl test catches these)

jchauncey: Yeah I cought it last ngiht. My next update to this PR will include more changes so I'll wait until then to fix it.

jchauncey:
Moving this back to in progress so I can add the cli upload job.

jobs/k8s_claimer.groovy
```diff
@@ -0,0 +1,300 @@
...
+
+      export DEIS_REGISTRY=quay.io/
+      docker login -e="\$QUAY_EMAIL" -u="\$QUAY_USERNAME" -p="\$QUAY_PASSWORD" quay.io
+
+      DOCKER_BUILD_FLAGS="--pull --no-cache" \
+      KUBECONFIG=kubeconfig \
+      ARGS=config.ssh_key=\${K8S_CLAIMER_SSH_KEY},\
+      config.google.account_file=\${GOOGLE_CLOUD_ACCOUNT_FILE},\
+      config.google.project_id=deis-e2e-leasable,\
+      config.auth_token=\${AUTH_TOKEN},\
```
sgoings, line 272: Will these show up in output?

jchauncey: It should not. We already do this for things like logging into the docker user for QUAY and Dockerhub

mboersma (approved):

PR #386, opened 2017-05-23
fix(jobs): Remove the CLOUD_PROVIDER from credentails binding

jchauncey:
this time....

### DIFF
--- a/jobs/apptypes_test.groovy
+++ b/jobs/apptypes_test.groovy
@@ -39,7 +39,6 @@ job(name) {
     colorizeOutput 'xterm'
     credentialsBinding {
       string("AUTH_TOKEN", "a62d7fe9-5b74-47e3-9aa5-2458ba32da52")
-      string("CLOUD_PROVIDER", defaults.e2eRunner.provider)
     }
   }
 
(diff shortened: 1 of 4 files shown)


jobs/apptypes_test.groovy
```diff
@@ -39,7 +39,6 @@ job(name) {
     colorizeOutput 'xterm'
     credentialsBinding {
       string("AUTH_TOKEN", "a62d7fe9-5b74-47e3-9aa5-2458ba32da52")
-      string("CLOUD_PROVIDER", defaults.e2eRunner.provider)
```
vdice, line 42: oh right, `defaults.e2eRunner.provider` is just a string.  However, if it is in a `credentialsBinding` wrapper, the value needs to be an actual credentials ID.  (Jenkins was treating `defaults.e2eRunner.provider`/`azure` as the literal credentials ID for looking up the actual secret to inject into the `CLOUD_PROVIDER` env var.)

(If we just wanted to set `CLOUD_PROVIDER` to `defaults.e2eRunner.provider`, we'd just wrap it in a `parameters{}` block in the form of a `stringParam()`, as seen elsewhere in this file...)

jchauncey: the last pr we merged moved it to being a stringparam but for some reason on my local machine the removal of this line was never staged into the commit. So this PR removes the lines which should have been removed on the last PR =P

vdice: Ah, and I see the latter is already done 👍

PR #390, opened 2017-05-31
fix(jobs/component_release.groovy): change to using repo name

vdice:


mboersma:
We're trying to do a release today, so I'm going to kick this to next month's milestone. @jchauncey PTAL when time permits.

jchauncey (approved):

